PR #18, opened 2026-03-07
Upgrade MLP to parameter-matched SwiGLU architecture

dumko2001:
### Architectural Upgrade
This PR replaces the standard Squared-ReLU `MLP` block with a modern SwiGLU architecture (`F.silu(fc1) * fc2`), which has become the empirical standard in SOTA transformer architectures (Llama 3, Mistral) due to better feature scaling and downstream validation loss.

### Fair Baseline Constraint
To ensure this is a strictly fair architectural upgrade and not just a parameter-count cheat, the hidden dimension is recalculated using an `8/3` multiplier against `n_embd`. This mathematically guarantees the total parameter count remains identical to the original `4x` expansion factor.

Additionally, the calculated hidden dimension is rounded up to the nearest multiple of 64 (`(hidden_dim + 63) // 64 * 64`) to ensure native alignment for Hopper Tensor Core utilization, avoiding any performance penalties from unaligned matrix multiplications.

### DIFF
--- a/train.py
+++ b/train.py
@@ -100,14 +100,18 @@ class CausalSelfAttention(nn.Module):
 class MLP(nn.Module):
     def __init__(self, config):
         super().__init__()
-        self.c_fc = nn.Linear(config.n_embd, 4 * config.n_embd, bias=False)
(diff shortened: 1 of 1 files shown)


karpathy:
What are you doing?

PR #45, opened 2026-03-08
Add multi-platform support (Jetson / SDPA fallback)

### LINKED ISSUE #5: MFU calculation is hardcoded to H100 peak FLOPS
## Problem

After PR #2 added support for non-Hopper GPUs, the MFU (Model FLOPS Utilization) metric is now misleading on any GPU that isn't an H100.

The peak FLOPS value is hardcoded at line 491:
\\\python
H100_BF16_PEAK_FLOPS = 989.5e12
\\\

On an A100 (312 TFLOPS BF16), the reported MFU would be ~3.2x too high. On an RTX 4090 (165 TFLOPS), it would be ~6x too high. This makes the MFU metric useless for comparing runs across different hardware.

## Proposed Fix

Auto-detect the GPU via \	orch.cuda.get_device_capability()\ and look up the correct theoretical peak FLOPS from a table of known GPU architectures. Fall back to the H100 value for unknown GPUs.

This is a minimal change (~10 lines) that makes the logging output accurate on all supported hardware.

I'll submit a PR for this.

hholtmann:
Gets autoresearch running on Jetson (and other non-H100 GPUs) without touching the H100 code path.

**What changed:**
- Attention: FA3 → FA2 → SDPA, picked automatically based on compute capability
- MFU: measures actual GPU peak FLOPS via matmul instead of hardcoding H100 value (fixes #5)
- Low VRAM (<16GB): activation checkpointing, smaller batches, bf16 softcap
- torch.compile skipped when Triton isn't available
- Single pyproject.toml: `uv sync` for H100 (unchanged), `uv sync --extra jetson` for Jetson

**Tested on Jetson Orin Nano Super** (7.4GB, sm_87): val_bpb 1.874, MFU 66.7%, 26 steps in 5 min.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ description = "Autonomous pretraining research swarm"
 readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
-    "kernels>=0.11.7",
+    "kernels>=0.11.7; platform_machine != 'aarch64'",
     "matplotlib>=3.10.8",
     "numpy>=2.2.6",
     "pandas>=2.3.3",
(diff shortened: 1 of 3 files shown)


svlandeg (commented):
Hi! Considering the readme

> People can reference (or have their agents reference) the full/parent nanochat repository that has wider platform support and shows the various solutions (e.g. a Flash Attention 3 kernels fallback implementation, generic device support, autodetection, etc.)

I assume we'll want to keep this particular train script clean and simple - I suggest to add these edits to your personal fork 😉

PR #52, opened 2026-03-08
Add cache integrity checks with weights_only and hash verification

### LINKED ISSUE #41: Harden cache artifact trust boundary in prepare.py
## Problem

Right now `prepare.py `trusts cached local artifacts too much when loading them again. In particular, it reloads files like `tokenizer.pkl `and `token_bytes.pt `directly from cache without any integrity check. That is usually fine for normal local use, but it becomes risky on shared machines, copied cache directories, or if cached files are tampered with.

## Proposed Fix

A fix would be to make cached artifact loading safer by using the safer `torch.load(..., weights_only=True) `path where possible, and adding a simple hash/integrity check before reusing cached tokenizer artifacts. This keeps the repo minimal while improving trust in cached files.

I will go ahead to submit a PR for this.

hkc5:
Fixes #41

This PR adds cache integrity checks for cached artifacts (tokenizer.pkl, token_bytes.pt) to address the security and reliability concerns raised in issue #41.

## Changes

1. **Added hash verification**: When saving tokenizer and token_bytes, a SHA256 hash is computed and stored alongside the file. When loading, the hash is verified to detect corruption.

2. **Added try/except for tokenizer loading (line 219)**: Wrapped pickle.load() in a try/except block that catches common pickle-related errors (UnpicklingError, EOFError, AttributeError, ImportError) and provides a helpful error message directing users to regenerate the cache.

3. **Added weights_only=True for torch.load (line 251)**: Changed torch.load() to use weights_only=True for safer deserialization of the token_bytes tensor.

## Benefits

- **Security**: Prevents potential code execution from malicious pickle files on shared machines
- **Reliability**: Detects cache corruption early with clear error messages
- **Backward compatibility**: If no hash file exists, the loader continues normally (graceful degradation)

## Testing

The changes maintain backward compatibility - existing caches without hash files will continue to work. New caches will have hash files generated automatically.

### DIFF
--- a/prepare.py
+++ b/prepare.py
@@ -15,6 +15,7 @@ import time
 import math
 import argparse
 import pickle
+import hashlib
 from multiprocessing import Pool
 
 import requests
@@ -174,9 +175,11 @@ def train_tokenizer():
         special_tokens=special_tokens,
     )
 
-    # Save tokenizer
-    with open(tokenizer_pkl, "wb") as f:
-        pickle.dump(enc, f)
+    # Save tokenizer with hash
+    def _save_tokenizer(enc, path):
+        with open(path, "wb") as f:
+            pickle.dump(enc, f)
+    _save_with_hash(enc, tokenizer_pkl, _save_tokenizer)
 
     t1 = time.time()
     print(f"Tokenizer: trained in {t1 - t0:.1f}s, saved to {tokenizer_pkl}")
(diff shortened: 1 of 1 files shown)


hkc5:
Closing - accidentally included changes from another PR. Will resubmit cleanly if needed.

PR #63, opened 2026-03-09
fix: replace pickle with JSON for tokenizer serialization

### LINKED ISSUE #41: Harden cache artifact trust boundary in prepare.py

johnwaldo:
Closes #41

## Problem

`prepare.py` uses `pickle.dump`/`pickle.load` for tokenizer serialization. Pickle deserialization executes arbitrary Python code ([CWE-502](https://cwe.mitre.org/data/definitions/502.html)), so anyone with write access to `~/.cache/autoresearch/tokenizer/tokenizer.pkl` (shared machines, copied caches, compromised filesystems) can achieve full code execution when `train.py` imports the tokenizer.

Additionally, `torch.load()` is called without `weights_only=True`, which relies on the PyTorch version default for safety — fragile across versions and forks.

## Changes

1. **Replace pickle with JSON** for tokenizer serialization. The `tiktoken.Encoding` object is reconstructed from its components (`pat_str`, `mergeable_ranks`, `special_tokens`) which are all JSON-serializable. Byte keys in `mergeable_ranks` are base64-encoded.

2. **Add `weights_only=True`** to `torch.load()` for `token_bytes.pt`, making the safe behavior explicit regardless of PyTorch version.

3. **Backward compatible**: If only the old `tokenizer.pkl` exists (no `tokenizer.json`), it still loads but prints a warning recommending re-running `prepare.py`. This means existing caches work without breakage.

## What this does NOT change

- No changes to `train.py`, the model, optimizer, or evaluation
- No new dependencies (uses stdlib `json` and `base64`)
- No changes to the tokenizer training logic itself — only the serialization format
- Existing users just need to re-run `python prepare.py` to regenerate the cache as JSON

## Note on formatting

A few cosmetic PEP 8 changes (line length wrapping, blank lines between top-level definitions) were applied by the editor alongside the security changes. The semantic diff is limited to the import block, `train_tokenizer()` save logic, `Tokenizer.from_directory()` load logic, and the `torch.load()` call.

### DIFF
--- a/prepare.py
+++ b/prepare.py
@@ -13,8 +13,9 @@ import os
 import sys
 import time
 import math
+import json
+import base64
 import argparse
-import pickle
 from multiprocessing import Pool
 
 import requests
@@ -27,8 +28,8 @@ import torch
 # Constants (fixed, do not modify)
 # ---------------------------------------------------------------------------
 
-MAX_SEQ_LEN = 2048       # context length
-TIME_BUDGET = 300        # training time budget in seconds (5 minutes)
+MAX_SEQ_LEN = 2048  # context length
+TIME_BUDGET = 300  # training time budget in seconds (5 minutes)
 EVAL_TOKENS = 40 * 524288  # number of tokens for val eval
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


johnwaldo:
Closing — not pursuing further. Reopen if useful.

Issue #47, opened 2026-03-08
improvements to novelty


ngoiyaeric:
how does adding agents ultimately guarantee a new architecture?

mkemka:
One approach I am experimenting with is to have two sub-agents with different backgrounds debate the best strategy to adopt.  This doesn't guarantee a new architecture but adds novelty.

ngoiyaeric:
so how do you measure the utility of novelty?

mkemka:
Currently I can only talk to the experiments I made in the fork (https://github.com/mkemka/autoresearch/blob/master/spiritualguidance.md).  There are two competing agents that argue and generate a combined directive that is used to alter the program.md for the next run. The history is stored in the spiritual guidance.md and used as a working memory. So to actually measure the utility I would need to see if there is actually novelty or variance of ideas from this approach and if in the long term the loss is lower compared to a single agent.

ngoiyaeric:
https://github.com/karpathy/autoresearch/pull/70 we can also do these manually like the novelty verification part you're referring too/ Seems to be an infinite loop.

ngoiyaeric:
you just added a readme, maybe @karpathy can chime in

wangyunjeff:
> Currently I can only talk to the experiments I made in the fork (https://github.com/mkemka/autoresearch/blob/master/spiritualguidance.md). There are two competing agents that argue and generate a combined directive that is used to alter the program.md for the next run. The history is stored in the spiritual guidance.md and used as a working memory. So to actually measure the utility I would need to see if there is actually novelty or variance of ideas from this approach and if in the long term the loss is lower compared to a single agent.目前我只能讨论我在分支代码（ https://github.com/mkemka/autoresearch/blob/master/spiritualguidance.md ）中做的实验。实验中有两个相互竞争的智能体，它们会进行争论并生成一个组合指令，该指令用于修改下次运行的 program.md 文件。历史记录存储在 spiritual guidance.md 文件中，并用作工作记忆。因此，要真正衡量这种方法的效用，我需要观察这种方法是否真的具有新颖性或思路多样性，以及从长远来看，其损失是否比单个智能体更低。

I've read your code; it's very interesting.Could you tell me exactly how to run the different agents?Is it within a single Codex instance or by starting separate instances?

ngoiyaeric:
[...]

PR #70, opened 2026-03-09
Fix training loss accumulation and add LR warmup

ngoiyaeric:
New Features

Added comprehensive training infrastructure with GPT-like model architecture, optimized optimizers, and integrated training loop with gradient accumulation support.
Improvements

Adjusted learning rate warmup schedule (now 5% instead of 0%).
Enhanced loss tracking during gradient accumulation for more accurate training metrics.


svlandeg (commented):
Not sure we'll want to merge this, but either way I'd remove the two new files?

ngoiyaeric:
true, this an experiment from this discussion https://github.com/karpathy/autoresearch/issues/47

PR #84, opened 2026-03-09
fix NaN loss not caught by fast-fail check

haosenwang1018:
The fast-fail check `train_loss_f > 100` silently passes on NaN because IEEE 754 NaN comparisons always return False. When an agent experiment produces NaN (e.g. from an aggressive LR change), the run wastes the full 5-minute budget training on garbage instead of aborting immediately.

`not (x <= 100)` is the standard negated-comparison idiom that catches both >100 and NaN, with no added complexity.

### DIFF
--- a/train.py
+++ b/train.py
@@ -9,6 +9,7 @@ os.environ["PYTORCH_ALLOC_CONF"] = "expandable_segments:True"
 os.environ["HF_HUB_DISABLE_PROGRESS_BARS"] = "1"
 
 import gc
+import math
 import time
 from dataclasses import dataclass, asdict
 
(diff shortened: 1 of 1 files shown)


svlandeg (approved):
Oh interesting, I didn't know this about direct `nan` comparisons! But I can confirm that `if not math.nan <= 100` would hit the fast fail, while `if math.nan > 100` wouldn't. So this fix is valid.

But wouldn't it be nicer & less error-prone to just check explicitely?
```
if train_loss is math.nan or train_loss > 100:
```

haosenwang1018:
Great suggestion — updated to an explicit check now:\n\n\n\nAlso added  accordingly.

haosenwang1018:
(Follow-up) Code snippet got stripped in previous comment by shell quoting. The update is now: if math.isnan(train_loss_f) or train_loss_f > 100.

PR #92, opened 2026-03-09
AgentHub

karpathy:
Call for help/discussion on autoresearch integration into [AgentHub](https://github.com/karpathy/agenthub). I have an early version deployed on [autoresearchhub.com](http://autoresearchhub.com/). The new `program.md` I am using for my first agent is below.

After some iteration I might push to master of autoresearch. Just want to iterate on first a bit more and think it through a bit.

### DIFF
--- a/program_agenthub.md
+++ b/program_agenthub.md
@@ -1,0 +1,191 @@
+
+# autoresearch — agenthub edition
+
+You are an autonomous research agent. You modify `train.py` to improve a language model's validation loss (`val_bpb`, lower is better). Each experiment runs for a fixed 5-minute time budget. You share your work through a central hub where multiple agents collaborate.
+
+## Hub API
+
+The hub is at `HUB=http://autoresearchhub.com`. All authenticated endpoints require `Authorization: Bearer <api_key>`.
+
+### One-time setup: register
+
+Credentials are stored in `~/.agenthub_creds`. If the file exists, you're already registered — just load it. Otherwise, register a new agent:
+
+```bash
+if [ -f ~/.agenthub_creds ]; then
+  source ~/.agenthub_creds
+else
+  # Pick a unique agent name and register
+  RESP=$(curl -s -X POST "$HUB/api/register" \
+    -H "Content-Type: application/json" \
+    -d '{"id":"YOUR_AGENT_NAME"}')
+  echo "$RESP"
+  # Returns: {"id":"...","api_key":"..."}
+  # Save credentials for future sessions
+  API_KEY=$(echo "$RESP" | python3 -c "import sys,json; print(json.load(sys.stdin)['api_key'])")
+  AGENT_ID=$(echo "$RESP" | python3 -c "import sys,json; print(json.load(sys.stdin)['id'])")
+  echo "export HUB_KEY=\"$API_KEY\"" > ~/.agenthub_creds
+  echo "export AGENT_ID=\"$AGENT_ID\"" >> ~/.agenthub_creds
+  source ~/.agenthub_creds
+fi
+```
+
+Use `$HUB_KEY` in all subsequent curl calls as `-H "Authorization: Bearer $HUB_KEY"`.
+
+### Git operations
+
+**Push a commit** (after a successful experiment):
+```bash
+git bundle create /tmp/push.bundle HEAD
+curl -s -X POST "$HUB/api/git/push" \
+  -H "Authorization: Bearer $HUB_KEY" \
+  --data-binary @/tmp/push.bundle
+```
+
+**Fetch a commit** (to build on someone else's work):
+```bash
+curl -s "$HUB/api/git/fetch/<hash>" \
+  -H "Authorization: Bearer $HUB_KEY" \
+  -o /tmp/fetch.bundle
+git bundle unbundle /tmp/fetch.bundle
+git checkout <hash>
+```
+
+**List recent commits**:
+```bash
+curl -s "$HUB/api/git/commits?limit=20" -H "Authorization: Bearer $HUB_KEY"
+```
+
+**Get frontier** (leaf commits — the tips of exploration with no children yet):
+```bash
+curl -s "$HUB/api/git/leaves" -H "Authorization: Bearer $HUB_KEY"
+```
+
+**Get children of a commit** (what's already been tried on top of it):
+```bash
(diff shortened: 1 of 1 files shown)


dumko2001:
@karpathy One question that came to mind while reading the experiment loop: the system is elegantly optimized for incremental hill-climbing: agents propose a change, it improves val_bpb, gets pushed, becomes part of the lineage. Clean and effective.
But some of the biggest jumps in ML historically required temporarily worse performance before unlocking a better regime : architectural overhauls, different scaling tradeoffs, training dynamics that look broken before they stabilize. The path to a higher peak sometimes means stepping into the valley first. And an agent that hard-discards anything below the current best has no way to take that step, it can only ever optimize the hill it's already standing on.
Humans handle this partly through intuition ,we can look across the landscape and sense that a taller mountain exists somewhere, even before the numbers confirm it. Agents here don't have that mechanism yet.
Have you thought about how to encourage that kind of exploration? A few directions that come to mind:

Letting agents maintain side branches that are temporarily worse but flagged as speculative
A separate exploration budget for more radical architectural changes
Agents occasionally sampling outside the current frontier rather than always building on the current best lineage

Otherwise I wonder if the system converges strongly to local optima over time — not because agents are bad, but because the incentive structure only rewards the next incremental step.
Curious how you're thinking about the exploration/exploitation tradeoff. Happy to dig into this more if it's something you're actively considering.

bigsnarfdude:
repo was deleted and forks point to https://github.com/ygivenx/agenthub

autonull:
multiobjective model optimization: max accuracy, min parameters, min iteration time
https://github.com/autonull/bioplausible/blob/83bfde7bf4469a97d4fb890569d9233db3d0577d/bioplausible/hyperopt/optuna_bridge.py#L212

dhanaway:
Autoresearch and AgentHub made me think a lot to evolution and a gene pool; it feels like there are shared characteristics between the two. In a gene pool there is no single 'main' branch lots of tracks are going at once in different directions each trying to find some new, better path. Similar to the vision of AgentHub, there is sharing and swapping of genes between these tracks (analogous to the sharing and swapping of commits on these branches). There is also no notion of a 'merge back into main' each track is going independently, some will fail and end, others will continue on and become the de facto 'best track' for a time.

I wonder if there is some design that builds on this proven strategy? Or maybe this is just an indication that the current design is a good one.

morozow:
@karpathy I propose to consider using **stdio Bus** – NDJSON over stdio, MCP/ACP-compatible – as the primary inter-agent routing base, instead of coordinating agent-to-agent traffic via AgentHub HTTP API calls.

Right now AgentHub coordinates work by explicit API endpoints – agents poll/post/claim through the hub. That hard-codes "where to send/how to coordinate" into the hub's API surface.

With stdio Bus, we can move coordination one level down: **routing is done by the agents themselves** over a deterministic stdio transport that is protocol-semantics-agnostic – MCP/ACP messages, or any custom JSON-RPC/NDJSON frames. In this model, the hub stops being a message router and becomes mainly:
- agents registry/discovery – who exists, capabilities
- durable storage for results, artifacts, experiment graph if desired
- optional policy/accounting

But the **actual message passing and task handoff** becomes: "agent/user/etc. > stdio Bus > agent".

Net effect:
- agents can dynamically choose peers, build their own topology – centralized, mesh, hybrid – and evolve routing strategy during runtime;
- transport stays unified NDJSON/MCP/ACP across local processes and remote workers.
- no need to encode routing decisions into REST endpoints;

I already started implementing this direction in [https://github.com/karpathy/autoresearch/pull/223](https://github.com/karpathy/autoresearch/pull/223) where the goal was to replace the HTTP-based hub coordination for the common case by local multi-GPU/local cluster with `stdio_bus` as the transport/router, including multi-GPU scheduling + parallel execution + sync across agents.

If this direction fits the goals of [AgentHub](https://github.com/karpathy/agenthub), I can sketch how the hub would expose only minimal "bootstrap + registry", while all inter-agent coordination runs through stdio Bus as the router/data-plane.

---

**References:**
- Website & Documentation: https://stdiobus.com/
- stdio Bus Kernel Repository: https://github.com/stdiobus/stdiobus
- Worker Registry: https://github.com/stdiobus/workers-registry

dexhunter:
> I wonder if there is some design that builds on this proven strategy? Or maybe this is just an indication that the current design is a good one.

@dhanaway I think most ai scientist with evolutionary algorithms were built upon this idea, maybe take a look at AlphaEvolve, [openevolve](https://github.com/algorithmicsuperintelligence/openevolve/tree/main), [shinkaevolve](https://github.com/SakanaAI/ShinkaEvolve) or [projects on the mle-bench leaderboard](https://github.com/openai/mle-bench)

mikeapedia:
@karpathy I just joined it this afternoon and I think it's good to go. My only question is if you plan to keep the python/pytorch/cuda versions fixed throughout the experiment or if you are open to letting the agents optimize that as well?
The system was easy to set up and flexible enough that I was able to build my own researcher loop using evoX (https://arxiv.org/pdf/2602.23413)

mikeapedia:
@karpathy 
image

harshitgavita-07:
Hey @karpathy ,

 This AgentHub + autoresearch direction is wild, thanks for opening it up for discussion. Reading through the loop and the comments here made me think about a slightly more “next‑gen” layer on top of the current design.

Right now agents optimize within a fixed program.md, proposing changes that improve val_bpb and propagating the best lineage. What if there was a dedicated meta‑agent whose job is to periodically analyze the entire experiment graph (all runs, configs, lineages) and actually edit the research program itself?

Concretely, the meta‑agent could:
– detect stale or low‑yield directions and gradually downweight/retire those branches based on longer‑horizon metrics, not just the latest score;
– propose new “research intents” (e.g., robustness, inference latency, data curation quality) and spin up scoped sub‑programs/agents to pursue them;
– re‑allocate GPU budget between exploration vs exploitation using a simple bandit/evolutionary signal over branches, instead of a fixed schedule.

This turns AgentHub into “agents that rewrite the research strategy” rather than just “agents that tweak the model,” and feels aligned with the evolution/gene‑pool analogies people mentioned above. I can prototype a minimal version of this meta‑agent (just reading the current experiment graph, emitting edits to program.md, and logging how it re‑allocates budget) and open a PR if that sounds useful.

Curious if this matches how you’re imagining the next layer of AgentHub, or if you see a cleaner place in the architecture for this kind of self‑reflective coordinator

Anurag1:
@Anurag1...push

PR #94, opened 2026-03-09
feat: Test Time RL - Auto Research

ar0cket1:
## Summary
- adds a thin TTT-Discover integration on top of autoresearch
- keeps the original autoresearch workflow intact
- uses the outer RL loop to propose `train.py` edits and uses inner-loop `val_bpb` as reward
- This should hypothetically (based on literature), massively outperform any frontier model, even when using TTT-Discover on a small OS model.

## Notes
- this has now been mostly tested and is confirmed to work in large part end to end
- I am currently still debugging it locally, but it works almost in entirety
- the main remaining rough edge is that the agent's code-edit replacement behavior can still be somewhat unreliable in some batches
- the most recent pushes are specifically aimed at improving that edit reliability

## Safety
- this PR does not overwrite upstream `master`
- it is proposed from a separate fork branch for review


ar0cket1:
*The replacement behaviour is fixed, though agents seem to be making edits that result in OOM errors, so nudging the prompt towards less OOM errors is the most recent fix.

PR #104, opened 2026-03-10
docs: fix README typos and style (macOS, copy-paste, grammar)

dhunganapramod9:
- Fix "a three files" → "three files"
- Fix "256 possibly bytes" → "possibly 256 bytes"
- Use macOS (Apple spelling) in Notable forks

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,13 +2,13 @@
 
 ![teaser](progress.png)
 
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
Thanks! I don't think the formatting changes are super necessary, and the typo is already fixed in #160 🙏

Issue #7, opened 2026-03-07
Training results lost if evaluation crashes (no pre-eval checkpoint)


NJX-njx:
## Problem

The training loop runs for the full 5-minute time budget, then calls \evaluate_bpb()\ to compute the val metric. If evaluation crashes (e.g., OOM when the agent has increased model size significantly, or a CUDA error), the entire 5-minute training run is lost with no recovery path.

This is especially problematic in the autonomous agent loop where experiments run overnight  a crash during eval wastes a full experiment cycle.

## Proposed Fix

Save a lightweight checkpoint (model state dict only) right after the training loop finishes and before evaluation begins. If evaluation succeeds, clean it up. If the process crashes during eval, the checkpoint survives for inspection or metric recovery.

Minimal change (~5 lines), no impact on the training loop itself.

I'll submit a PR for this.

PR #112, opened 2026-03-10
Fix Issue #7: Save checkpoint before evaluation to prevent training loss

teee32:
## Summary

- **Fixes Issue #7**: Training results lost if evaluation crashes (no pre-eval checkpoint)

### Problem
When running unattended overnight experiments, if evaluation crashes (OOM, CUDA errors, etc.) after the full 5-minute training budget, the entire training run is lost.

### Solution
- Save checkpoint before each evaluation
- Retry failed evaluations up to 3 times with smaller batch size
- Progressive wait time (2s, 4s, 6s) between retries for GPU memory release
- GPU memory diagnostics for debugging
- Auto-cleanup checkpoint after successful evaluation

### Changes
- `train.py`: +217 lines (checkpoint utilities, retry logic, memory management)

### Test plan
- [ ] Run evaluation that triggers OOM and verify checkpoint is saved
- [ ] Verify retry mechanism works with smaller batch size
- [ ] Verify checkpoint is cleaned up after successful evaluation

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ By design, training runs for a **fixed 5-minute time budget** (wall clock, exclu
 
 ## Quick start
 
-**Requirements:** A single NVIDIA GPU (tested on H100), Python 3.10+, [uv](https://docs.astral.sh/uv/).
+**Requirements:** A single NVIDIA GPU with CUDA access (tested on H100), Python 3.10+, [uv](https://docs.astral.sh/uv/).
 
 ```bash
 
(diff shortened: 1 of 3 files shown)


Roy3838:
Hey! This overnight problem is exactly why I built a free/FOSS tool (I do quantum physics simulations :D )
https://github.com/Roy3838/Observer 
It's a tiny local agent that watches your screen and pings or calls you the second something crashes (OOM, CUDA errors, whatever), so you're not finding out the run was lost only after the full budget burned. 
Inference local, nothing leaves your machine. 
Happy to answer anything :D

PR #136, opened 2026-03-10
Add AMD ROCm support via PyTorch SDPA fallback

bkpaine1:
## Summary

- Auto-detect Flash Attention 3 availability at import; gracefully fall back to PyTorch native `F.scaled_dot_product_attention` when FA3/kernels are unavailable (AMD ROCm, CPU, or missing kernel packages)
- SDPA fallback correctly handles GQA (grouped query attention) head expansion via `repeat_interleave`
- Replace hardcoded `H100_BF16_PEAK_FLOPS` with `GPU_BF16_PEAK_FLOPS` — auto-detects AMD devices and allows override via `GPU_BF16_PEAK_FLOPS` env var for accurate MFU reporting on any hardware
- Update README requirements to include AMD ROCm alongside NVIDIA

**Zero behavior change on NVIDIA** — the FA3 path is unchanged when kernels are available. The try/except only fires when FA3 import fails.

## Motivation

The autoresearch loop is hardware-agnostic by design — the agent experiments with architecture, not drivers. But the training script currently hard-requires NVIDIA FA3 kernels, locking out the growing AMD ROCm ecosystem (MI300X, consumer Radeon, Strix Halo APUs). PyTorch's native SDPA is well-optimized on ROCm and provides a clean fallback without sacrificing much performance.

## Testing

- Syntax verified
- SDPA fallback path tested with autoresearch's exact tensor shapes (B=2, T=64, n_head=6, head_dim=128)
- Both standard attention (n_kv_head == n_head) and GQA (n_kv_head < n_head) configurations pass
- Output shapes match FA3 path exactly

## Changes

- `train.py`: 47 lines added, 10 removed
- `README.md`: Requirements updated, ROCm note added

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,9 @@ If you are new to neural networks, this ["Dummy's Guide"](https://x.com/hooeem/s
 
 ## Quick start
 
-**Requirements:** A single NVIDIA GPU (tested on H100), Python 3.10+, [uv](https://docs.astral.sh/uv/).
+**Requirements:** A single GPU (NVIDIA or AMD ROCm), Python 3.10+, [uv](https://docs.astral.sh/uv/).
+
+> **AMD ROCm support:** The training script automatically falls back to PyTorch's native SDPA when Flash Attention 3 is unavailable. Install [PyTorch for ROCm](https://pytorch.org/get-started/locally/) and run as normal. Set `GPU_BF16_PEAK_FLOPS` env var for accurate MFU reporting on your hardware.
 
 ```bash
 
--- a/train.py
+++ b/train.py
@@ -16,11 +16,20 @@ import torch
 import torch.nn as nn
 import torch.nn.functional as F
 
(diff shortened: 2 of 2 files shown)


Bortlesboat:
This also fixes Windows users hitting the same FA3 issue. On Windows, `kernels-community/flash-attn3` fails with:

```
FileNotFoundError: Kernel at path ... does not have one of build variants: torch29-cu128-x86_64-windows, torch-cuda, torch-universal
```

The try/except approach here catches it correctly. One remaining Windows-specific blocker not covered by this PR: `torch.compile` requires Triton which is Linux-only. On Windows you additionally need `backend="eager"` on all `torch.compile` calls, or a similar platform check.

Tested this SDPA approach on RTX 3060 12GB / Windows 11 / CUDA 12.9 and it runs correctly — val_bpb 1.42 in 5 min.

PR #151, opened 2026-03-10
Add optional checkpoint import/export hooks

corysmart:
## Summary

This adds optional checkpoint load/save hooks to `train.py` without changing default behavior.

New behavior is only enabled when these env vars are set:

- `AUTORESEARCH_LOAD_CHECKPOINT`
- `AUTORESEARCH_SAVE_CHECKPOINT`

## What changed

- add `maybe_load_checkpoint()` to load a checkpoint before training
- allow partial state-dict restore when tensor shapes still match
- add `maybe_save_checkpoint()` to persist model state and a small amount of run metadata after evaluation
- keep the compiled training path unchanged by loading/saving around the existing `model_core` setup

## Why

This creates a minimal continuation/export seam that is useful for:
- resuming or bootstrapping experiments
- external harnesses that want to persist model state between short runs
- downstream tooling that needs checkpoint artifacts without restructuring the trainer

## Notes

- default behavior is unchanged when the env vars are unset
- checkpoint loading is permissive by shape, so compatible subsets can still be restored
- no new dependencies were added

### DIFF
--- a/train.py
+++ b/train.py
@@ -11,6 +11,7 @@ os.environ["HF_HUB_DISABLE_PROGRESS_BARS"] = "1"
 import gc
 import time
 from dataclasses import dataclass, asdict
+from pathlib import Path
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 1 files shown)


mvanhorn:
Similar problem space - I opened #265 as a minimal take on this. Single `torch.save` before eval, 3 lines, no env vars. Thought it was worth having a lightweight option too.

PR #166, opened 2026-03-11
Fix notebook undefined best in analysis notebook

### LINKED ISSUE #262: Bug: 'best' variable used before definition in analysis.ipynb

SamuelS0:
Line 137 references 'best' on lines 51-52, but 'best' was never defined, causing a NameError when running cells top-to-bottom. Fixed by assigning 'best = running_min.iloc[-1]' which is already computed earlier in the same cell.


svlandeg (commented):

PR #170, opened 2026-03-11
Fix MLX optimizer KeyError on dict-keyed parameters

elementalcollision:
## Summary
- Fixed `KeyError: 1` crash in `_set_param_by_path` when navigating `value_embeds` (a dict with string keys like `"0"`, `"1"`)
- The old code assumed all numeric path segments were list indices (`obj[int(part)]`), but dicts need string keys (`obj[part]`)
- Added `_navigate_part` helper that checks `isinstance(obj, dict)` to dispatch correctly
- Added MLX training run screenshot to README

## Test plan
- [x] Verified MLX training starts successfully on M1 Max (previously crashed at step 1)
- [ ] Full 5-minute training run completes without errors


elementalcollision:
Opened against upstream by mistake — moving to fork.

PR #193, opened 2026-03-11
Enhance experiment runner with deterministic controls

buzypi:
## Summary

This PR adds a execution workflow for autonomous experiments, replacing session-by-session `program.md` interpretation with a deterministic runner plus agent runbook.

### What’s included

- Add `workflows/run_experiment.py` as the single experiment orchestrator:
  - `start`, `resume`, `status` commands
  - top-level stage controls: `setup`, `baseline`, `loop`
  - loop sub-stage controls: `propose`, `apply`, `commit`, `train`, `triage`, `record`, `decide`
  - resumable checkpointing under `workflows/runs/<run_id>/`
  - run-id policy: `<branch-slug>-rNNN`
- Add `AGENTS.md` runbook with explicit natural-language to command mapping for agent sessions.
- Setup robustness:
  - auto-run `uv run prepare.py` when cache/tokenizer is missing (default on, opt-out via `--no-auto-prepare`)
  - explicit setup precondition checks before baseline/loop
- Background training support:
  - training stages start in background by default (`--background-train`)
  - `resume` polls/continues in-flight baseline/train jobs
- Human intervention support:
  - proposal override via run-scoped `workflows/runs/<run_id>/next_proposal.json`
  - proposal override via explicit `--proposal-file <path>` on `start`/`resume`
  - canonical override schema at `workflows/schemas/proposal.schema.json`
  - deterministic precedence in propose stage: `--proposal-file` -> `next_proposal.json` -> stochastic proposal -> deterministic fallback
  - consumed run-scoped proposals are archived to `workflows/runs/<run_id>/consumed_proposals/iter_<NNNN>.json`

## Why

In long-running autonomous sessions, prose-only execution is fragile and inconsistent. Multiple complaints regarding this on X.
This PR makes runs repeatable, resumable, inspectable, and human-steerable while preserving `program.md` as the policy/objective layer.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,96 @@
+# Agent Runbook: Autoresearch Execution
+
+Use `workflows/run_experiment.py` for all autoresearch execution. Do not use `workflow/dag.json` or task-graph-planner for running experiments.
+
+## Core Rules
+
+1. Keep all run artifacts under `workflows/runs/`.
+2. Modify only `train.py` during experiments.
+3. Never modify `prepare.py`.
+4. Always run training via `uv run train.py` through the script.
+
+## Natural-language to Command Mapping
+
+- User says: "Start running the experiment, run 5 loops"
+  - Run: `python workflows/run_experiment.py start --loops 5`
+
+- User says: "Run another 5 iterations"
+  - Run: `python workflows/run_experiment.py resume --loops 5`
+
+- User says: "Resume run <run_id> and run 5 loops"
+  - Run: `python workflows/run_experiment.py resume --run-id <run_id> --loops 5`
+
+- User says: "Only run setup and baseline"
+  - Run: `python workflows/run_experiment.py start --only setup,baseline`
+
+- User says: "Only run training and decision parts in loops for 3 iterations"
+  - Run: `python workflows/run_experiment.py resume --loops 3 --only loop --loop-only train,record,decide`
+
+- User says: "Show run status"
+  - Run: `python workflows/run_experiment.py status`
+
+## Stage Controls
+
+- Top-level stages: `setup`, `baseline`, `loop`
+- Loop stages: `propose`, `apply`, `commit`, `train`, `triage`, `record`, `decide`
+
+Supported control flags:
+
+- `--only <comma-list>`: run only selected stages
+- `--from-stage <setup|baseline|loop>` + `--to-stage <...>`: run a top-level stage range
+- `--loop-only <comma-list>`: limit loop internals to selected stages
+- `--loops N`: run `N` loop iterations
+
+## Human Proposal Override
+
+- You can inject a human-authored proposal for the next `propose` stage.
+- Option A (run-scoped default): write JSON to `workflows/runs/<run_id>/next_proposal.json`.
+- Option B (explicit path): pass `--proposal-file <path>` on `start`/`resume`.
+- Proposal JSON must include exactly these keys:
+  - `status` (`ok` or `need_input`)
+  - `description`
+  - `change_plan`
+  - `commit_description`
+- Canonical schema: `workflows/schemas/proposal.schema.json`
+- Example:
+  ```json
+  {
+    "status": "ok",
+    "description": "Try slightly higher learning rate with warmup unchanged.",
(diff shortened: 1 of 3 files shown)


buzypi:
## Basic End-to-End Flow (via OpenCode)

1. Start OpenCode in repo root:

```bash
opencode
```

2. Ask it to start a run:

Type this in OpenCode: "Start running the experiment, run 5 loops"

It maps to:

```bash
python workflows/run_experiment.py start --loops 5
```

3. Continue later:

Type this in OpenCode: "Run another 5 iterations"

It maps to:

```bash
python workflows/run_experiment.py resume --loops 5
```

4. Check status:

The training runs happen in the background. And the agent goes into a sleep (which it auto-determines). We can interrupt it and ask it questions like: "Show the run status".

It maps to:

```bash
python workflows/run_experiment.py status
```

We can ask any other questions like: "Tell me about the results so far" or "What is the GPU usage" etc. We can ask it to resume its work.

## Human-in-the-Loop Override

If you want to inject your own experiment idea instead of stochastic proposal generation:

Type this in OpenCode: "In the next iteration, increase the LR by 10% and keep warmup unchanged. Use this as a human proposal and run 1 loop."

## Useful Stage-Control Examples (Direct)

In case you want to go headless you can do these:

Setup + baseline only:

```bash
python workflows/run_experiment.py start --only setup,baseline
```

Run only selected loop internals for 3 iterations:

```bash
python workflows/run_experiment.py resume --loops 3 --only loop --loop-only train,record,decide
```

Foreground mode (disable background training):

```bash
python workflows/run_experiment.py resume --loops 1 --no-background-train
```

## Logs and Artifacts

Run outputs are written to:

- `workflows/runs/<run_id>/runner.log` (human-readable timeline)
- `workflows/runs/<run_id>/history.jsonl` (structured events)
- `workflows/runs/<run_id>/state.json` (checkpoint state)

morozow:
@buzypi PR introduces a Python runner that orchestrates the loop via explicit scripted stages – start, resume, status and propose, either apply, train, and other.

I'm proposing a different abstraction: move orchestration from "scripted control-flow" to "transport-level routing". With [stdio Bus](https://github.com/karpathy/autoresearch/pull/223), agents are processes connected via stdin/stdout, and coordination happens by routing NDJSON messages over stdio – MCP/ACP or any custom protocol — no endpoint-driven coordination, no hub-defined call graph.

Bus interconnection becomes the orchestration engine: it routes, supervises, and multiplexes agent-to-agent messages at the stdio level.

buzypi:
I think the 2 solve very different problems. The main issue that this PR targets is to be able to still achieve what @karpathy was trying via program.md but make the loop more robust and not rely on LLMs. @karpathy didn't like a Ralph Loop because it does not allow the ability for human in the loop interaction. So the loop implementation here, while being deterministic, is invoked via an agent harness (currently OpenCode). The agent gets to know about the script via AGENTS.md. So we still have the natural language control over the execution. We can interrupt the execution, ask for stats, even suggest plans. But when we say, "Loop forever" or "Loop 100 times" it will follow the instructions more deterministically. That's all is the purpose of this PR.

The PR you suggested is for distributed GPU training and coordination which this PR does not address.

Further, the Design Choices clearly mention:

Self-contained. No external dependencies beyond PyTorch and a few small packages. **No distributed training**, no complex configs. **One GPU**, one file, one metric.

morozow:
Yep, agreed – we're solving different problems. https://github.com/karpathy/autoresearch/pull/193 is about making the loop deterministic, restartable and less dependent on LLMs for execution. My comment was more of a "long-range" direction: moving orchestration onto a transport-protocol layer, so any feature (e.g., benchmarks, training, evaluation, tooling) plugs into the same runtime substrate. The execution is still deterministic and fully resumable because the worker and runtime enforces the state machine, checkpointing, and replay semantics. The difference is that control and orchestration flow through the protocol – message routing over stdio – not through a separate either CLI or scripts-driven implementation.

Also, the "infinite loop" concern is different in this model, it's not fire-and-forget. The loop stays fully controllable in real time – we can interrupt, redirect, re-plan, or change conditions mid-run, because workers operate against live files, DB, queues state and can be updated continuously while the run is ongoing.

PR #201, opened 2026-03-12
[BUG] fix: notebook miss value 'best'

### LINKED ISSUE #262: Bug: 'best' variable used before definition in analysis.ipynb

yuyu5333:
# Motivation
Before the pr, an error would occur when drawing "progress.png" because the 'best' variable was missing.

# Modification
Added the best variable before drawing the image.

# Bug info
image


svlandeg (commented):

PR #203, opened 2026-03-12
Introduce Component-System for component optimization, add a dashboar…

LaurenceLong:
…d to monitor activities


LaurenceLong:
Wrong operation

PR #232, opened 2026-03-13
fix: skip torch.compile on Windows

Bortlesboat:
## Summary

Skip `torch.compile` on Windows so `train.py` can run when Triton-backed compilation is unavailable.

This keeps the current behavior on non-Windows platforms, but treats compilation as optional on Windows by routing both the fused optimizer helpers and the model compile call through a small `maybe_compile(...)` wrapper.

## Why

I reproduced this on Windows 11 with a local RTX 3060 using the repo's existing environment:

```text
os.name= nt
torch= 2.9.1+cu128
cuda= True
device= NVIDIA GeForce RTX 3060
compile_fail TritonMissing Cannot find a working triton installation.
```

Right now that prevents `train.py` from starting unless the user manually edits out the compile calls.

## Scope

- leave Linux / non-Windows behavior unchanged
- make the compile path a no-op on Windows
- avoid broad backend / platform changes

## Validation

- `python -m py_compile train.py`
- reproduced `torch.compile(...)` failure in the local Windows env
- verified the same code path succeeds when wrapped with `maybe_compile(...)`

### DIFF
--- a/train.py
+++ b/train.py
@@ -25,6 +25,15 @@ fa3 = get_kernel(repo).flash_attn_interface
 
 from prepare import MAX_SEQ_LEN, TIME_BUDGET, Tokenizer, make_dataloader, evaluate_bpb
 
+
+def maybe_compile(target=None, **kwargs):
+    """Skip torch.compile on Windows where Triton-backed compilation is unavailable."""
+    if target is None:
+        return lambda real_target: maybe_compile(real_target, **kwargs)
+    if os.name == "nt":
+        return target
+    return torch.compile(target, **kwargs)
+
 # ---------------------------------------------------------------------------
 # GPT Model
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


Bortlesboat:
Repro details from the local Windows box I tested on:

```text
os.name= nt
torch= 2.9.1+cu128
cuda= True
device= NVIDIA GeForce RTX 3060
compile_fail TritonMissing Cannot find a working triton installation.
```

This is from the repo's existing Windows 11 environment. Without this patch, `train.py` requires manually editing out the `torch.compile(...)` calls before it will start.

morozow:
@Bortlesboat A similar issue is addressed in https://github.com/karpathy/autoresearch/pull/230 (see [line 322](https://github.com/karpathy/autoresearch/pull/230/changes#diff-ed183d67207df065a11e1289f19d34cc2abbc5448dea952683cfe9728c342b95R322)). We plan to minimize the resolver’s dependency footprint.

Bortlesboat:
Closing this to avoid duplicating the backend-direction already being handled in #230. My original Windows repro is still real on current `master`, but given the maintainers plan to fold this into the broader backend work while keeping the resolver surface tighter, I would rather step out than leave a redundant PR open.

PR #263, opened 2026-03-14
fix: define 'best' variable before use in progress plot

### LINKED ISSUE #262: Bug: 'best' variable used before definition in analysis.ipynb

mango766:
Fixes #262. \n\nThe variable `best` is used to calculate the margin and set the y-axis limits in the progress plot, but it is not defined in the scope of that cell. It is defined in a later cell. I've added the definition before its use to prevent `UnboundLocalError`.


svlandeg (commented):

PR #153, opened 2026-03-10
feat(autoresearch): add checkpoint, config, and queue management system

BenBreaksIn:
- Add checkpoint.py for model save/load with resume capability
- Add config.py for centralized configuration with environment variable overrides
- Add gpu_queue.py for file-based GPU job queue and time-sharing across agents
- Add knowledge.py for agent knowledge base and context management
- Add launch.py for experiment orchestration and agent coordination
- Add run_experiment.py for end-to-end experiment execution
- Add programs/ directory with agent role definitions (analyst, director, explorer, optimizer, reviewer)
- Update .gitignore to exclude checkpoints/ and *.log files
- Update analysis.ipynb to support expanded results.tsv schema with additional metrics
- Update prepare.py and train.py for integration with new checkpoint and config systems
- Update program.md with comprehensive system documentation
- Enables multi-agent autoresearch with GPU sharing, checkpointing, and structured experimentation

PR #265, opened 2026-03-14
save checkpoint before eval to survive crashes

### LINKED ISSUE #7: Training results lost if evaluation crashes (no pre-eval checkpoint)

mvanhorn:
if evaluate_bpb() OOMs or hits a CUDA error, the entire training run is lost. save model state_dict before eval so the agent can recover.

closes #7

### DIFF
--- a/train.py
+++ b/train.py
@@ -606,6 +606,9 @@ print()  # newline after \r training log
 
 total_tokens = step * TOTAL_BATCH_SIZE
 
+# Save checkpoint before eval so training isn't lost if eval crashes
+torch.save(model.state_dict(), "checkpoint.pt")
+
 # Final eval
 model.eval()
 with autocast_ctx:


mvanhorn:
Good point - #151 adds 58 lines with env var hooks and partial state-dict restore, #153 adds 2230 lines with a full checkpoint/config/queue system. This PR is intentionally just one `torch.save` call, 3 lines, no new config surface. The repo's design choices section emphasizes "self-contained, no complex configs" - this tries to match that spirit.

svlandeg:
Erm, no, the PRs you reference aren't the ones linked to #7 😉

mvanhorn:
Ha, whoops - wrong PR numbers, sorry abut that. The ones actually linked to #7 are #14 (+10 lines) and #112 (+507/-207 lines). This PR takes the same approach as #14 but pared down to the bare minimum - just the `torch.save` call, 3 lines total.

PR #282, opened 2026-03-15
Bake reflection into the experiment loop

alexbenari:
## Summary

This updates `program.md` to add a lightweight reflection step before and after each experiment.

The workflow now asks the agent to:
- initialize a `musings.md` file during setup
- write down a short rationale for the proposed idea, including the underlying intuition and any relevant ML grounding, before changing `train.py`
- record a brief post-mortem after the run, capturing whether the attempt succeeded and what was learned.

The resulting `musings.md` leaves behind a useful learning trail for humans (at least this human) interested in the nuances of ML theory and their application.
It may also improve the agent's idea generation quality by making the loop more systematic, though that would need more deliberate validation.

## Changes

- add `musings.md` initialization to the experiment setup checklist
- require a short pre-execution writeup for each proposed idea
- require a short post-run reflection on whether the attempt worked and what was learned

## Testing

- docs/process change only

### DIFF
--- a/program.md
+++ b/program.md
@@ -14,7 +14,8 @@ To set up a new experiment, work with the user to:
    - `train.py` — the file you modify. Model architecture, optimizer, training loop.
 4. **Verify data exists**: Check that `~/.cache/autoresearch/` contains data shards and a tokenizer. If not, tell the human to run `uv run prepare.py`.
 5. **Initialize results.tsv**: Create `results.tsv` with just the header row. The baseline will be recorded after the first run.
-6. **Confirm and go**: Confirm setup looks good.
+6. **Initialize musings.md**: Create an empty `musings.md`. You will be writing to it before each idea you try
+7. **Confirm and go**: Confirm setup looks good.
 
 Once you get confirmation, kick off the experimentation.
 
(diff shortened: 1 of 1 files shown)


morozow:
@alexbenari A memorizable text file for the agent is a good idea. Do not forget to instruct the agent to read this file before starting. A prompt-engineered `musings.md` context provides more valuable sense for agent than a large amount of historical data alone. 

I experimented with a historical file-based context approach in my fork of autoresearch repo, highlighted in discussion https://github.com/karpathy/autoresearch/discussions/234, and I agree that historical file-system context works well.

alexbenari:
It might be useful for resuming agent work as you suggest, although for that purpose musings.md specifically seems somewhat of an overkill. A much smaller file would do (perhaps even the existing `results.tsv`). 
I think the main value of musings.md is as 1. a learning aid for the human 2. steering the agent's reasoning

morozow:
1. learning aid for the human – the only practical option here is a `musings.md` collection/DB, or similar store. Nobody is going to read every LLM experimental runtime outcome
3. steering the agent's reasoning – is essentially the same as "useful for resuming agent work as you suggest"

Nevertheless, memorizable text file for agent is still a good idea.

PR #306, opened 2026-03-17
Clarify experiment bookkeeping in program.md

Gabrieliam42:
## Summary
- clarify that `train.py` is the only source file to edit
- explicitly allow `results.tsv` bookkeeping and `run.log` updates as experiment artifacts
- require the `results.tsv` row to be appended before any reset, checkout, or revert

## Why
The previous wording could be read as forbidding the log and results bookkeeping that the experiment loop itself requires. This makes the intended workflow explicit without broadening what source files may be changed.

### DIFF
--- a/program.md
+++ b/program.md
@@ -11,7 +11,7 @@ To set up a new experiment, work with the user to:
 3. **Read the in-scope files**: The repo is small. Read these files for full context:
    - `README.md` — repository context.
    - `prepare.py` — fixed constants, data prep, tokenizer, dataloader, evaluation. Do not modify.
-   - `train.py` — the file you modify. Model architecture, optimizer, training loop.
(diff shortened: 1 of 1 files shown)


elementalcollision:
Technically, program.md is also a source file to edit, as you can designate which dataset you want to experiment on.  I don't necessarily view this as ambiguous as even the agent (in my case Opus and/or Sonnet) understands which files are immutable and which aren't.

Gabrieliam42:
All right,
It's just that when I ran it with Codex on VS Code it didn't update `results.tsv` or `run.log` because it interpreted that train.py directive as a override restriction, so it ignored the other directives, therefore it didn't work as intended with Codex and GPT-5.4, probably Claude is more permissive. After I modified it as stated it began to work as the project intended.

PR #310, opened 2026-03-17
code optimization

qingfeng123-www:
Change the dataset path to the root directory of the project to prevent occupying C drive space. All parameter configurations in train.py have been placed at the forefront for easy modification. I Use backup links to download datasets to prevent network issues from causing dataset download failures.


qingfeng123-www:
code optimization

PR #331, opened 2026-03-18
write results.json for structured agent consumption, harden crash diagnostics

### LINKED ISSUE #64: Indirect prompt injection via training output fed back to agent

shichangs:
## Summary

Fixes #64

- `train.py`: writes `results.json` after evaluation with the same metrics already printed to stdout. Gives agents a structured, parseable results channel instead of relying on grepping free-form stdout from `run.log`.
- `program.md`: instructs agent to read `results.json` first (fallback to grep). Crash diagnostics now use filtered `grep -i "error|exception|traceback"` instead of raw `tail -n 50`, reducing the surface for indirect prompt injection via training output.
- `.gitignore`: adds `results.json` and `run.log` as runtime artifacts.

Existing stdout output is unchanged. No new dependencies (`json` is stdlib).

Similar to #79 but rebased on current main with no merge conflicts.

## Test plan

- [ ] Run `uv run train.py` and verify `results.json` is written with correct metrics
- [ ] Verify stdout output is unchanged
- [ ] Verify `results.json` values match stdout summary
- [ ] Simulate a crash (e.g. syntax error in train.py) and confirm `results.json` is NOT written

### DIFF
--- a/train.py
+++ b/train.py
@@ -26,6 +26,10 @@ fa3 = get_kernel(repo).flash_attn_interface
 
 from prepare import MAX_SEQ_LEN, TIME_BUDGET, Tokenizer, make_dataloader, evaluate_bpb
 
+# Clear previous results so a killed run leaves no file
+if os.path.exists("results.json"):
+    os.remove("results.json")
+
 # ---------------------------------------------------------------------------
 # GPT Model
 # ---------------------------------------------------------------------------


shichangs:
cc @johnwaldo (issue author) @mvanhorn (PR #79 author) — would appreciate your review. This takes a similar approach to #79 but rebased cleanly on current main.

### Offline verification

The `results.json` output can be validated without a GPU:

```python
import json

# Simulate what train.py writes
results = {
    "val_bpb": 0.9979,
    "training_seconds": 300.1,
    "total_seconds": 325.9,
    "peak_vram_mb": 45060.2,
    "mfu_percent": 39.8,
    "total_tokens_M": 499.6,
    "num_steps": 953,
    "num_params_M": 50.3,
    "depth": 8,
}
with open("results.json", "w") as f:
    json.dump(results, f, indent=2)

# Verify roundtrip
with open("results.json") as f:
    loaded = json.load(f)
assert loaded == results
assert isinstance(loaded["val_bpb"], float)
assert isinstance(loaded["num_steps"], int)
print("✓ results.json roundtrip OK")
```

The key behavioral changes:
1. **Success path**: agent reads `results.json` (structured JSON) instead of grepping stdout → no injection surface
2. **Crash path**: agent uses `grep -i "error|exception|traceback"` to filter relevant lines before falling back to `tail` → reduced injection surface

shichangs:
Thanks for the thorough review @johnwaldo!

1. **Stale `results.json`** — Great catch. Fixed in 4318f43: `results.json` is now deleted at the top of `train.py` before training starts, preserving the "no file = crash" contract.

2. **Crash-path grep injection surface** — Agreed this is mitigation, not elimination. A structured `error.json` on the crash path would be the complete fix, but that feels like a separate PR. Leaving as-is for now.

3. **PR #79 cleanup** — Will close #79 with a pointer here.

@mvanhorn Closing #79 in favor of this one — same approach, rebased on current main. Thanks for the original work there!

mvanhorn:
Nice rebase. The crash diagnostics filter is a good addition over my original #79. Closing mine in favor of this.

shichangs:
@karpathy This PR is ready for review — all feedback from johnwaldo has been addressed, and #79 has been closed in favor of this one. Small change (3 files), happy to make any further adjustments if needed.

PR #373, opened 2026-03-22
Support training on GB10 (Spark DGX) without flash-attn/torch.compile

ramdurga:
## Summary
- Auto-detect GPU capabilities (SM version) and conditionally use flash-attn3 and torch.compile
- GPUs outside PyTorch's supported SM 8.0–12.0 range (e.g. GB10 at SM 12.1) fall back to PyTorch SDPA and eager mode
- No change for supported GPUs (H100, A100, etc.) — they still get flash-attn3 + torch.compile

## Test plan
- [x] Verified training runs to completion on NVIDIA GB10 (SM 12.1) — val_bpb 1.87, 26 steps, loss 9.01→5.48
- [ ] Verify no regression on Hopper/Ampere GPUs (flash-attn + torch.compile path unchanged)

### DIFF
--- a/train.py
+++ b/train.py
@@ -17,11 +17,24 @@ import torch
 import torch.nn as nn
 import torch.nn.functional as F
 
-from kernels import get_kernel
 cap = torch.cuda.get_device_capability()
-# varunneal's FA3 is Hopper only, use kernels-community on non-Hopper GPUs
-repo = "varunneal/flash-attention-3" if cap == (9, 0) else "kernels-community/flash-attn3"
-fa3 = get_kernel(repo).flash_attn_interface
+# flash-attn3 has CUDA kernels for SM 8.0-12.0 only
+if (8, 0) <= cap <= (12, 0):
+    try:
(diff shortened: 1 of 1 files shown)


ukd1:
btw you can make this work on GB10 differently if you want close to FA - its not that much worse (still is) compared to before

 pyproject.toml:
  - Removed kernels dependency
  - torch 2.9.1 → 2.10.0
  - CUDA 12.8 → CUDA 13.0

  train.py:
  - Replaced kernels/FA3 imports with flex_attention + create_block_mask from torch.nn.attention
  - Changed attention forward pass: FA3's flash_attn_func(q, k, v, causal=True, window_size=...) → flex_attention(q, k, v, block_mask=..., enable_gqa=True) with explicit BHtd transpose
  - Added _precompute_block_masks() method that creates FlexAttention block masks encoding both causal + sliding window constraints
  - window_size tuples replaced with precomputed block_mask objects throughout the call chain (forward → Block → CausalSelfAttention)

PR #426, opened 2026-03-27
fix(tokenizer): resolve prepend_id scope issue in encode()

Agolid:
## Summary

Fix `NameError` and `ValueError` in `Tokenizer.encode()` when `prepend` is a string.

## Problem

1. `prepend_id` is assigned inside `if prepend is not None:` but referenced in a nested block — if `encode_single_token()` raises, `prepend_id` is never defined
2. `encode_single_token()` raises `ValueError` for multi-token prepend strings
3. Inner guard checks `prepend` instead of `prepend_id`, masking failed computation

## Fix

- Initialize `prepend_id = None` before conditional blocks
- Wrap `encode_single_token()` in try/except with fallback to `encode_ordinary()`
- Check `prepend_id is not None` instead of `prepend is not None` for insertion

## Changes

```python
# Before: prepend_id defined only in inner scope
if prepend is not None:
    prepend_id = prepend if isinstance(prepend, int) else self.enc.encode_single_token(prepend)

# After: proper initialization and error handling
prepend_id = None
if prepend is not None:
    try:
        prepend_id = prepend if isinstance(prepend, int) else self.enc.encode_single_token(prepend)
    except ValueError:
        prepend_id = self.enc.encode_ordinary(prepend)[0] if prepend else None
```

Fixes #421

### DIFF
--- a/prepare.py
+++ b/prepare.py
@@ -226,15 +226,20 @@ class Tokenizer:
         return self.bos_token_id
 
     def encode(self, text, prepend=None, num_threads=8):
+        prepend_id = None
         if prepend is not None:
-            prepend_id = prepend if isinstance(prepend, int) else self.enc.encode_single_token(prepend)
+            try:
+                prepend_id = prepend if isinstance(prepend, int) else self.enc.encode_single_token(prepend)
+            except ValueError:
+                # Fallback: encode as regular text if prepend is multi-token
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
Duplicate of #421, closing.

PR #493, opened 2026-04-07
fix: use raw token bytes for BPB calculation

### LINKED ISSUE #384: Bug: BPB metric inflated by UTF-8 replacement characters in token byte count
### Description

In `prepare.py` (lines 188-193), the token byte count for BPB evaluation is computed by decoding each token via tiktoken and re-encoding to UTF-8:

```python
token_str = enc.decode([token_id])
token_bytes_list.append(len(token_str.encode("utf-8")))
```

For BPE tokens whose raw bytes are not valid standalone UTF-8 (e.g. a single continuation byte `0x80` = 1 raw byte), `tiktoken.decode()` replaces invalid sequences with U+FFFD (the Unicode replacement character), which encodes to 3 UTF-8 bytes. This inflates the byte-count denominator in BPB, producing **artificially lower (better-looking) scores**.

### Impact

The BPB metric — the core evaluation metric that drives the entire autoresearch loop — is silently underestimated. Experiments may appear to perform better than they actually do, and comparisons between runs with different tokenizers (which may have different ratios of non-UTF-8 tokens) are not apples-to-apples.

### Fix

Use `mergeable_ranks` directly to get the true raw byte length of each token:

```python
rank_to_bytes = {rank: raw for raw, rank in mergeable_ranks.items()}
for token_id in range(enc.n_vocab):
    if token_id in rank_to_bytes:
        token_bytes_list.append(len(rank_to_bytes[token_id]))
    else:
        token_bytes_list.append(0)  # special tokens
```

I will submit a fix shortly.

MohammadWasi:
Replace tiktoken decode/encode approach with direct mergeable_ranks lookup to avoid UTF-8 replacement
character inflation in evaluation metrics.

The old method could inflate BPB scores when BPE tokens contained invalid UTF-8 sequences, as tiktoken.decode() replaces them with U+FFFD (3 bytes) instead of the actual raw byte length (often 1 byte).

Fixes #384

### DIFF
--- a/prepare.py
+++ b/prepare.py
@@ -185,11 +185,17 @@ def train_tokenizer():
     print("Tokenizer: building token_bytes lookup...")
     special_set = set(SPECIAL_TOKENS)
     token_bytes_list = []
+    # Create reverse lookup: rank -> raw bytes for accurate byte counting
+    rank_to_bytes = {rank: raw for raw, rank in mergeable_ranks.items()}
+    
     for token_id in range(enc.n_vocab):
-        token_str = enc.decode([token_id])
-        if token_str in special_set:
+        if token_id in special_tokens.values():
             token_bytes_list.append(0)
+        elif token_id in rank_to_bytes:
+            token_bytes_list.append(len(rank_to_bytes[token_id]))
         else:
+            # This should not happen, but fallback for safety
+            token_str = enc.decode([token_id])
             token_bytes_list.append(len(token_str.encode("utf-8")))
     token_bytes_tensor = torch.tensor(token_bytes_list, dtype=torch.int32)
     torch.save(token_bytes_tensor, token_bytes_path)


MohammadWasi:
## Problem
The BPB (bits-per-byte) evaluation metric was artificially inflated when BPE tokens contained invalid UTF-8 sequences. The current implementation uses [tiktoken.decode()](cci:1://file:///Users/mohammadwasi/Documents/open-source/autoresearch/prepare.py:249:4-250:35) which replaces invalid bytes with U+FFFD (replacement character), encoding to 3 UTF-8 bytes instead of the actual raw byte length (often 1 byte).

## Impact
This caused experiments to appear better than they actually were, potentially misleading the autonomous research agent and compromising the integrity of all evaluation metrics that drive the autoresearch loop.

## Solution
Replace the tiktoken decode/encode approach with direct `mergeable_ranks` lookup to get the true raw byte length of each token:

```python
# Before (buggy):
token_str = enc.decode([token_id])
token_bytes_list.append(len(token_str.encode("utf-8")))

# After (fixed):
rank_to_bytes = {rank: raw for raw, rank in mergeable_ranks.items()}
for token_id in range(enc.n_vocab):
    if token_id in rank_to_bytes:
        token_bytes_list.append(len(rank_to_bytes[token_id]))
    else:
        token_bytes_list.append(0)  # special tokens

svlandeg (commented):
Closing, duplicate of https://github.com/karpathy/autoresearch/pull/385

PR #510, opened 2026-04-11
refactor(train): wrap entrypoint in main() and fix CUDA alloc env var

mendentm:
- Rename PYTORCH_ALLOC_CONF to PYTORCH_CUDA_ALLOC_CONF so expandable segments are actually enabled (the previous name was silently ignored, leaving the allocator optimization off in V1/V2).
- Move setup, training loop, and final eval into a main() function under an `if __name__ == "__main__"` guard so train.py is importable for smoke tests, multi-seed runs, and tooling without executing on import.
- Lift get_lr_multiplier, get_muon_momentum, get_weight_decay, and build_model_config to module scope; build_model_config now takes vocab_size as an argument instead of closing over a module global.
- Drop the unused startup_time local.

Hyperparameter block and all model/optimizer code are byte-identical, so agent-tunable knobs and run-to-run diffs remain trivially reviewable.

### DIFF
--- a/train.py
+++ b/train.py
@@ -5,7 +5,7 @@ Usage: uv run train.py
 """
 
 import os
-os.environ["PYTORCH_ALLOC_CONF"] = "expandable_segments:True"
+os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
 os.environ["HF_HUB_DISABLE_PROGRESS_BARS"] = "1"
 
 import gc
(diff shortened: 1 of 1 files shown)


mendentm (commented):
New

svlandeg (commented):
> Rename PYTORCH_ALLOC_CONF to PYTORCH_CUDA_ALLOC_CONF so expandable segments are actually enabled (the previous name was silently ignored, leaving the allocator optimization off in V1/V2).

What version of pytorch are you on? `PYTORCH_CUDA_ALLOC_CONF` is actually deprecated, and `PYTORCH_ALLOC_CONF` is recommended in the [docs](https://docs.pytorch.org/docs/stable/notes/cuda.html#optimizing-memory-usage-with-pytorch-alloc-conf).

srosro:
Heads up — @svlandeg's correction here led me to verify on PyTorch 2.11, and the env-var rename in this PR (and #546, and my own closed-then-rewritten #588) is a workaround for what turned out to be a **PyTorch 2.9.1 bug**, not a karpathy typo. Setting `PYTORCH_ALLOC_CONF=expandable_segments:True` is silently ignored for CUDA in 2.9.1 but works correctly in 2.11.0. PyTorch issue: [pytorch/pytorch#183387](https://github.com/pytorch/pytorch/issues/183387) (closed with "fixed in 2.11.0" confirmation).

So the actual root-cause fix is bumping `torch==2.9.1` → `torch==2.11.0` in `pyproject.toml`. After the bump, the upstream `train.py:8` line works as written and this PR's rename becomes unnecessary (and triggers a per-run deprecation warning on 2.11+).

Two open PRs already make this bump:
- **#504** (comprehensive: 2.11 + FlexAttention + SDPA + MFU)
- **#588** (focused: 1-line `pyproject.toml` bump, just-rewrote it after this finding)

The `main()` wrap in this PR is separately valid hygiene and stands on its own — just flagging that the allocator part of the diff is now redundant with the version bump.

svlandeg (commented):
Thanks for the digging @srosro! To reduce noise, I'll go ahead and close this PR.

PR #516, opened 2026-04-15
feat: add macOS CPU/MPS support alongside CUDA

marketswitch:
## What changed

`pyproject.toml` hardcoded `torch` to the `pytorch-cu128` index, which only ships Linux and Windows wheels. This makes the project completely uninstallable on macOS.

This PR adds a platform-conditional torch source:

- **Linux / Windows** → `pytorch-cu128` (existing CUDA 12.8 behaviour, unchanged)
- **macOS** → `pytorch-cpu` (resolves a CPU+MPS wheel, uses Apple Silicon GPU via MPS)

The torch pin is also relaxed from `==2.9.1` to `>=2.11.0` so the CPU index can resolve a compatible wheel.

## Why

Developers on Macs couldn't even run `uv sync` to get started. With MPS available on Apple Silicon, local experimentation is viable — training is slow vs. H100 but works for iterating on ideas before running overnight jobs on a GPU server.

## Testing

```bash
uv sync
uv run python -c "import torch; print(torch.__version__); print(torch.backends.mps.is_available())"
# torch: 2.11.0
# MPS available: True
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,15 +13,21 @@ dependencies = [
     "requests>=2.32.0",
     "rustbpe>=0.1.0",
     "tiktoken>=0.11.0",
-    "torch==2.9.1",
+    "torch>=2.11.0",
 ]
 
 [tool.uv.sources]
 torch = [
-    { index = "pytorch-cu128" },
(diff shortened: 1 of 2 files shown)


stabgan:
Is this tested properly ? very interesting

PR #540, opened 2026-04-23
Fix insecure deserialization torch load 2033323798907486799

aniruddhaadak80:


svlandeg (commented):
11 opened PRs from the same user within 5 minutes of eachother. Closing. Please don't DOS attack OS repos :pray:

PR #542, opened 2026-04-24
Minor cleanups: README typo and redundant .tmp check in prepare.py

tempoo04:
Two small fixes:

- README: "possibly bytes" → "possible bytes"
- prepare.py: drop redundant .tmp filter — temp files end in .parquet.tmp, already excluded by the .parquet check.


349369236:
哈哈

349369236:
yoududu

PR #546, opened 2026-04-29
fix: PYTORCH_CUDA_ALLOC_CONF env var + scope shard-download response

lonexreb:
## Summary

Two small, focused fixes that don't overlap with any other open PR:

- **`PYTORCH_ALLOC_CONF` → `PYTORCH_CUDA_ALLOC_CONF`** in `train.py`. PyTorch silently ignores the previous name, so `expandable_segments:True` was never actually applied and the CUDA caching allocator ran with more fragmentation than intended. PR #510 also fixes this but bundles it with a larger function-level refactor; this PR is just the one-line rename so it's mergeable independently.
- **Scope `requests.get` to a `with`-block** in `prepare.py::download_single_shard`. `stream=True` requests leave the underlying urllib3 connection open until GC runs the response object's finalizer; with the parallel-download fan-out (`Pool(processes=8)`), sockets can pile up. Wrapping the response in `with` releases them deterministically on both success and partial-stream failures inside `iter_content`.

Both changes are commented inline so the rationale is visible to future readers.

## Test plan

- [x] `python -m py_compile train.py prepare.py`
- [ ] `uv run prepare.py` — confirms shard download still completes (no functional change in success path)
- [ ] `uv run train.py` — confirms training still starts; `nvidia-smi` should show fewer/no allocator warnings on tight VRAM configs once expandable segments engages

### DIFF
--- a/prepare.py
+++ b/prepare.py
@@ -65,13 +65,16 @@ def download_single_shard(index):
     max_attempts = 5
     for attempt in range(1, max_attempts + 1):
         try:
-            response = requests.get(url, stream=True, timeout=30)
-            response.raise_for_status()
-            temp_path = filepath + ".tmp"
-            with open(temp_path, "wb") as f:
-                for chunk in response.iter_content(chunk_size=1024 * 1024):
-                    if chunk:
-                        f.write(chunk)
+            # Scope the response to a with-block so the underlying socket is
+            # released promptly on success or partial-stream failure, instead
+            # of waiting for GC to trigger urllib3 cleanup.
+            with requests.get(url, stream=True, timeout=30) as response:
+                response.raise_for_status()
+                temp_path = filepath + ".tmp"
+                with open(temp_path, "wb") as f:
+                    for chunk in response.iter_content(chunk_size=1024 * 1024):
+                        if chunk:
+                            f.write(chunk)
             os.rename(temp_path, filepath)
             print(f"  Downloaded {filename}")
             return True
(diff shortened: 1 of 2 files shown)


srosro:
Heads up — verified on PyTorch 2.11 that the env-var rename in this PR (and #510, and my own closed-then-rewritten #588) is a workaround for what turned out to be a **PyTorch 2.9.1 bug**, not a karpathy typo. Setting `PYTORCH_ALLOC_CONF=expandable_segments:True` is silently ignored for CUDA in 2.9.1 but works correctly in 2.11.0. PyTorch issue: [pytorch/pytorch#183387](https://github.com/pytorch/pytorch/issues/183387) (closed with "fixed in 2.11.0" confirmation).

So the actual root-cause fix is bumping `torch==2.9.1` → `torch==2.11.0` in `pyproject.toml`. After the bump, the upstream `train.py:8` line works as written and this PR's rename becomes unnecessary (and triggers a per-run deprecation warning on 2.11+).

Two open PRs already make this bump:
- **#504** (comprehensive: 2.11 + FlexAttention + SDPA + MFU)
- **#588** (focused: 1-line `pyproject.toml` bump, just-rewrote it after this finding)

The `requests.get` context-manager scoping in this PR is separately valid hygiene and stands on its own — just flagging that the allocator part of the diff is now redundant with the version bump.

svlandeg (commented):
Thanks for the digging @srosro! To reduce noise, I'll go ahead and close this PR.

PR #555, opened 2026-04-30
feat(program): add Lévy-flight mutation-size guidance to the loop

lonexreb:
## Summary
A pure prompt change in ``program.md``. Before each experiment, the agent samples a mutation size from a heavy-tailed distribution and lets it constrain the change it proposes:

- **65%** *short* — single hyperparameter / one line
- **25%** *medium* — 2–3 related knobs
- **10%** *long*  — deliberately radical (swap optimizer family, replace activation, restructure residual path, …)

## Why
The current loop is a greedy hill-climber with effectively-uniform step sizes (every commit looks roughly the same in surface area). On a sparse-reward, non-convex landscape that is exactly the regime where Lévy-flight foragers — sharks, albatrosses, fruit flies, deer — measurably outperform Brownian-motion foragers. The same heavy-tailed perturbation operator powers the most-cited modern metaheuristics (Cuckoo Search, Manta Ray, Prairie Dog, …).

The 10% long-jump budget isn't claiming most long jumps win — most won't. It's the cheap insurance against the hill-climber stalling on a plateau separated from a better basin by even a small ridge.

## Why this is the smallest possible step
``bioinspired.md`` (PR #554) proposes 9 patterns ordered cheapest-first. This is pattern #1 — pure prompt change, zero infra, zero code, single paragraph in ``program.md``. It validates whether the proposal is actionable in isolation before any of the more invasive patterns (pheromone trail, replay buffer, lateral roots) are layered on.

## Citations
- [The Lévy flight foraging hypothesis](https://pubmed.ncbi.nlm.nih.gov/25709823/)
- [Survey of Lévy-flight metaheuristics, *Mathematics* 2022](https://www.mdpi.com/2227-7390/10/15/2785)
- [Manta Ray Foraging + chaotic Lévy + restart, *Sci. Rep.* 2025](https://www.nature.com/articles/s41598-025-25766-y)

## Test plan
- [x] Diff is +8 lines, all in one paragraph
- [x] Existing loop semantics unchanged — agent still hill-climbs, this only constrains *how big* the next step is
- [ ] If merged, observe over a multi-hour autoresearch run whether long jumps measurably extend the time-to-stagnation. Easy to A/B against a parallel run on a sibling branch without this paragraph.

PR #554, opened 2026-04-30
docs: bio-inspired search strategies for the autoresearch loop

lonexreb:
## Summary
``program.md`` describes the autoresearch loop as a single-track greedy hill-climber. It works, but it has well-known failure modes on sparse-reward, high-dimensional, non-convex search:

- no memory across runs (failed ideas can be re-tried, near-misses get forgotten)
- no explicit exploration / exploitation balance
- no global topology of the search space
- single track, even on multi-GPU setups

This signature — *sparse-reward, patchy, non-convex* — is exactly what foraging predators, root systems, slime molds, and the hippocampus have been pressured to solve. ``bioinspired.md`` proposes 9 patterns from those domains and maps each to a concrete, **incrementally adoptable** extension of the loop.

## What's in the proposal
The patterns are ordered cheapest-first (most are pure ``program.md`` prompt changes; only a few add infra):

| #  | Pattern                                  | Why it helps the loop                          |
| -- | ---------------------------------------- | ---------------------------------------------- |
| 1  | Lévy-flight perturbations                | escape local optima with rare radical jumps    |
| 2  | Pheromone trail with evaporation         | bias next-axis selection by per-axis success   |
| 3  | Slime-mold reinforcement / pruning       | turn parallel tracks into evolutionary tournament |
| 4  | Hippocampal replay + dopamine RPE        | renew variation from recent surprising wins    |
| 5  | Inhibition of return / novelty bonus     | stop thrashing the same corner of search space |
| 6  | Apical dominance + lateral roots         | save fallback hypotheses; break stalls cleanly |
| 7  | Cuckoo abandon-the-worst-nest            | self-correct multi-agent setups                |
| 8  | Phyllotaxis golden-angle sampling        | low-discrepancy sweeps over continuous knobs   |
| 9  | Octopus distributed cognition            | loose coupling beats tight sync at >1 GPU      |

Each section contains (a) the biology, (b) why it works, (c) the mapping to autoresearch, and (d) the smallest implementation — typically a paragraph in ``program.md`` plus an optional small TSV the agent maintains in plain text.

## Why this is a docs PR (and not a code PR)
``prepare.py`` is fixed; ``train.py`` is the experiment surface, not the search-loop surface. The search loop *is* ``program.md`` — a prompt that the agent executes. Most of these patterns are best added by editing the prompt, not by writing Python that wraps the agent. Shipping the proposal as a doc lets the project decide which mechanisms (if any) to fold into ``program.md`` itself.

## Citations
[...]

PR #558, opened 2026-04-30
feat(program): add inhibition-of-return axis-rotation rule

lonexreb:
## Summary
A pure prompt change in ``program.md``. Default behaviour: don't propose a change on the *same axis* as the immediately previous experiment — rotate to a different axis (LR vs head_dim vs optimizer vs attention window vs weight decay vs activation, …). The exception is a discard that left a targeted lesson on the same axis (\"LR=0.05 diverged → try LR=0.03\"); then keep going on that axis using the lesson as the explicit reason.

## Why
The agent's worst failure mode at scale is *thrashing* — many small variations on whichever knob it most-recently touched, because the latest result is the most salient piece of context. The primate visual system has the same problem when saccading and suppresses it with **inhibition of return**: a brief refractory period on the just-attended location that forces the next saccade to cover ground rather than re-fixate. The optimization analogue is the *novelty exploration bonus* used in deep RL agents like Random Network Distillation and Never Give Up.

Combined with the Lévy-flight mutation-size policy in PR #555, the agent gets:
- **broad axis coverage** (this PR)
- on top of **heavy-tailed step sizes** (PR #555)

That's the minimum exploration scaffolding before pheromone trails / replay buffers / lateral roots from ``bioinspired.md`` (PR #554) become worthwhile.

## Citations
- [Lilian Weng, Exploration Strategies in Deep RL](https://lilianweng.github.io/posts/2020-06-07-exploration-drl/)
- [Wittmann et al., novelty exploration bonus, *Neuropsychologia* 2009](https://www.sciencedirect.com/science/article/abs/pii/S0028393209000190)
- [Exploration in Deep RL: a survey, arXiv 2205.00824](https://arxiv.org/pdf/2205.00824)

## Test plan
- [x] Diff is +4 lines, all in one paragraph + a one-line carve-out for the exception case
- [x] Existing loop semantics unchanged — agent still hill-climbs, this only adds a soft preference for axis rotation
- [ ] Optional follow-up: log the touched axis as a 6th column in ``results.tsv`` so the rotation policy is auditable post-hoc
- [ ] If merged with PR #555, an A/B against a sibling ``autoresearch/<tag>-control`` branch without these two paragraphs would surface whether they measurably extend time-to-stagnation


svlandeg (commented):

grvwrk:
> I think all of these LLM-generated PRs are best submitted to your fork instead 🙏

brutal  😭

PR #559, opened 2026-04-30
docs: physics-inspired search strategies (companion to bioinspired.md)

lonexreb:
## Summary
Where ``bioinspired.md`` (#554) proposes new *mechanisms* for the autoresearch loop, this companion proposes *guardrails*. It is a first-principles audit of the loop in the voice and discipline of statistical mechanics, simulated annealing, and the Feynman ratchet. Eight patterns, ordered cheapest-first; the first five are pure protocol changes to ``program.md`` plus a few extra columns in ``results.tsv``.

## The patterns

| #  | Pattern                                         | What it protects against                                |
| -- | ----------------------------------------------- | ------------------------------------------------------- |
| 1  | **Measure the noise floor first**               | Celebrating BPB wiggles below the irreducible ~3e-4 floor |
| 2  | **Metropolis-Hastings acceptance**              | Greedy traps in a deceptive landscape                   |
| 3  | **Fluctuation-dissipation curvature**           | Constant-size steps regardless of local stiffness       |
| 4  | **Feynman ratchet check**                       | Accidental asymmetric reject rules ratcheting into noise |
| 5  | **Action principle: MDL prior**                 | Two agents arguing about \"simplicity\" in English      |
| 6  | **Renormalization-group scale-check**           | Wins that don't transfer to larger DEPTH                |
| 7  | **Honest loss / second-order signals**          | Wins that mask blown-up gradient norms or low attn entropy |
| 8  | **Path integrals over architectures (sketch)**  | The point you only need once #1–#5 are saturating       |

## Why this exists alongside bioinspired.md
A greedy ``T = 0`` descent on a noisy landscape has two distinct failure modes:
- **Insufficient exploration** — bio-inspired patterns (Lévy, IoR, pheromone, replay, …) all attack this.
- **False positives at the noise floor** — physics-inspired guardrails (noise floor, MDL, ratchet check, RG, second-order signals) attack this.

Both are needed. The bio doc gives you *what to do next*; the physics doc gives you *whether the thing you just did was real*.

## Calibrated, not aspirational
Each pattern has a back-of-envelope number, not just a vibe:
- The noise floor: ``σ_stat ≈ 0.5 / sqrt(21e6) / log(2) / 3 ≈ 5e-5`` from the val shard size; CUDA-kernel jitter pushes effective floor to ~3e-4.
- Metropolis cooling: ``T_0 ≈ 0.005``, geometric decay to ``T_end ≈ 1e-5``; gives a calibrated table of accept probabilities for typical ΔE.
[...]

PR #560, opened 2026-04-30
feat(program): measure-the-noise-floor protocol — "you are the easiest person to fool"

lonexreb:
## Summary
A pure prompt change in ``program.md``. After the first baseline run, do a *second* baseline run on the unchanged code. The two ``val_bpb`` values will differ — because of CUDA kernel non-determinism, GC scheduling jitter, and floating-point reduction order. Their absolute difference is the **noise floor σ_noise**.

Then every future experiment is interpreted against this floor:

- ``|Δval_bpb| < 2 × σ_noise`` → **discard, no matter the sign.** Indistinguishable from noise.
- ``|Δval_bpb| ≥ 2 × σ_noise`` → normal keep/discard rule applies.

## Why
Cargo Cult Science, Caltech 1974:
> *"The first principle is that you must not fool yourself — and you are the easiest person to fool."*

The current greedy rule (\"any improvement is a keep\") treats every BPB wiggle as signal. Back-of-envelope, the irreducible statistical floor on a 21M-token val shard with per-token loss std ~0.5 and ~3 bytes/token is

```
σ_stat  ≈  0.5 / sqrt(21e6) / log(2) / 3  ≈  5e-5
```

CUDA kernel jitter pushes the empirical floor to a few × 1e-4. Anything below ~3e-4 is structurally suspect. Without measuring it, the agent celebrates 0.0003-BPB \"wins\" and then anchors the next 12 experiments on what is, in expectation, a coin flip. One extra baseline run at the start of the session pays for itself the first time it catches a phantom KEEP.

This is pattern #1 from ``physics_inspired.md`` (PR #559) — the cheapest and highest-leverage protocol change. No code, no infra; pure ``program.md`` paragraph.

## How this stacks with prior PRs
- **PR #555 (Lévy-flight mutation size)** controls *step size*.
- **PR #558 (inhibition of return)** controls *which axis to step on*.
- **This PR** controls *whether the step actually happened* — i.e. whether to believe the metric reading.

Together: the agent gets broad axis coverage with heavy-tailed step sizes, and refuses to anchor on wiggles below the noise floor. That is the minimum exploration *and* honesty scaffolding before any of the more ambitious patterns from ``bioinspired.md`` / ``physics_inspired.md`` (pheromone trails, Metropolis acceptance, MDL prior, replay buffers) become worthwhile.

## Test plan
- [x] Diff is +8 lines, all in one paragraph + a 3-bullet decision rule
- [x] Existing loop semantics unchanged — agent still hill-climbs, this only adds a floor under the accept rule
- [ ] On the first session run with this prompt, observe what σ_noise actually is on H100 vs A100 vs other; the back-of-envelope ~3e-4 will be calibrated empirically the first time anyone runs the protocol.
- [ ] Optional: log σ_noise as a header comment in ``results.tsv`` for cross-session comparability

## Citations
- [Feynman, *Cargo Cult Science*, Caltech 1974](https://calteches.library.caltech.edu/51/2/CargoCult.htm)
- ``physics_inspired.md`` §1 in this repo (PR #559)

### DIFF
--- a/program.md
+++ b/program.md
@@ -38,6 +38,14 @@ Each experiment runs on a single GPU. The training script runs for a **fixed tim
 
 **The first run**: Your very first run should always be to establish the baseline, so you will run the training script as is.
 
+**Measure the noise floor (second run)**: After the first baseline run, do a *second* baseline run on the unchanged code. The two `val_bpb` values will differ because of CUDA kernel non-determinism, GC scheduling, and floating-point reduction order. The absolute difference between them is your **noise floor σ_noise** — back-of-envelope, ~3e-4 BPB on the 21M-token val shard, but measure it, don't assume it. Record both as `keep` (the second is the new baseline). All future experiments are interpreted against this floor:
+
+- |Δval_bpb| < 2 × σ_noise → **discard**, no matter the sign. The change is indistinguishable from noise; keeping it anchors 12+ downstream experiments on what is essentially a coin flip.
+- |Δval_bpb| ≥ 2 × σ_noise *and* improving → **keep** (subject to the simplicity criterion).
+- Anything else → **discard**.
+
+This is the single highest-leverage protocol change in the loop: most "0.0003 BPB improvements" celebrated in greedy hill-climbs are noise. One free run at the start saves dozens of experiments wasted building on a phantom win.
+
 ## Output format
 
 Once the script finishes it prints a summary like this:


lonexreb:
A falsifiable prediction so this PR isn't pure protocol — back-of-envelope for σ_noise on H100 with the default config:

The naive analysis using bf16 machine epsilon (`ε_bf16 ≈ 7.8 × 10⁻³`) gives a pessimistic ~0.19 BPB, which can't be right since the repo visibly converges to BPB ≈ 0.86. The reason it overshoots: bf16 matmuls on Hopper accumulate partial products in **fp32**, then round only the final output to bf16. So `ε_fp32 ≈ 1.2 × 10⁻⁷` is the reduction epsilon, not `ε_bf16`.

The actual run-to-run noise comes from **kernel scheduling non-determinism** — different valid orderings of partial products through fp32 accumulators give different fp32 totals, which then round to slightly different bf16 outputs. Modelling this as `c · ε_bf16` per matmul output with `c ~ 0.1` (only a small fraction of outputs sit near a bf16 round-off boundary):

```
per-matmul output relative noise          ~ 0.1 · ε_bf16        ≈ 7.8 × 10⁻⁴
per-token loss noise (absolute)           ~ 5 · 7.8e-4 · 2.0 nats ≈ 7.8 × 10⁻³ nats
val-mean loss noise after partial decorr. ~ 7.8e-3 / √10           ≈ 2.5 × 10⁻³ nats
σ_noise (BPB)                             = 2.5e-3 / log(2) / 3   ≈ 1.2 × 10⁻³ BPB
```

So I predict the actual observed σ_noise on H100 lands in the range **4 × 10⁻⁴ to 4 × 10⁻³ BPB**, central estimate ~10⁻³, dominated by FA3 kernel scheduling on the attention pass. (My original ~3 × 10⁻⁴ guess in physics_inspired.md was on the optimistic side — keeping it here for the record.)

If anyone runs the protocol I'd be curious whether this lands. The most informative outcome is the *spread* across runs, not the absolute number — if it's tightly bunched, scheduling non-determinism is well-controlled; if it's >2× variation, FA3 ordering is the dominant source and the protocol's `2σ` threshold needs to be fit empirically per-GPU rather than assumed universal.

The point of writing this down ahead of measurement: a bad prediction graded honestly is more useful than a hedged one. I'd rather find out my back-of-envelope is wrong than have it look right by being vague.

svlandeg (commented):

PR #564, opened 2026-04-30
feat(program): golden-angle sampling for continuous-knob sweeps

lonexreb:
## Summary
Adds one paragraph to ``program.md`` instructing the agent: *when sweeping a continuous hyperparameter (LR, weight decay, embedding scale, gate temperature, dropout, …), sample from a low-discrepancy golden-ratio sequence on log-space rather than uniform random or grid.*

The paragraph includes a 5-line stdlib-only snippet:

```python
GOLDEN = 0.61803398875
def golden_iter(n, lo, hi, seed=0.5):
    x = seed
    out = []
    for _ in range(n):
        x = (x + GOLDEN) % 1.0
        out.append(lo * (hi / lo) ** x)        # log-space interpolation
    return out
```

## Why
Plants distribute leaves at the [golden angle 137.5°](https://nph.onlinelibrary.wiley.com/doi/10.1111/nph.16040) because the irrationality of φ is *exactly* what guarantees no two leaves stack — every leaf gets clear sky access. The same fact makes [the generalized golden ratio one of the lowest-discrepancy quasi-random sequences](https://extremelearning.com.au/unreasonable-effectiveness-of-quasirandom-sequences/) we know.

At the budgets autoresearch operates on (5–30 candidates per sweep), this matters in practice:
- **Uniform random** clusters by chance — small samples have visible gaps.
- **Grid search** clusters at endpoints — half your candidates are at the boundary, where the gradient is usually small.
- **Golden-angle sampling** covers the space monotonically with each new candidate, so even a 5-point sweep has near-optimal spread.

Roberts (linked above) measures ~10–30% efficiency improvement over random at small budgets.

## Verified locally
```
candidates (8 LRs in [1e-4, 1e-1]): 0.0002, 0.0162, 0.0012, 0.0825, 0.0059, 0.0004, 0.0301, 0.0022
log-spread (max - min in log10):     2.56
```
Eight points, evenly distributed across 2.56 decades, no clustering, no boundary bias.

## Citations
- [Strauss et al., phyllotaxis golden angle for light capture, *New Phytologist* 2020](https://nph.onlinelibrary.wiley.com/doi/10.1111/nph.16040)
- [Mitchison, biophysical optimality of the golden angle, *Sci. Rep.* 2015](https://www.nature.com/articles/srep15358)
- [Roberts, *The unreasonable effectiveness of quasi-random sequences*](https://extremelearning.com.au/unreasonable-effectiveness-of-quasirandom-sequences/)
- [GoldenSequences.jl reference implementation](https://github.com/mschauer/GoldenSequences.jl)
- ``bioinspired.md`` §8 in this repo (PR #554)

## Independence
- Pure prompt change — no new files, no new dependencies (the snippet uses stdlib only).
- Independent of all other open PRs. Composes naturally with PR #555 (Lévy mutation size): a *medium* mutation that picks an LR is exactly the case where you want golden sampling instead of "I'll try 0.05".

## Test plan
- [x] Diff is +19 lines, all in one paragraph + the 5-line snippet
- [x] Snippet is stdlib-only Python — no torch, no numpy required
- [x] Verified output: 8 candidates evenly distributed across 2.56 log-decades for ``golden_iter(8, 1e-4, 1e-1)``
- [ ] Optional follow-up: add a small ``tools/golden_iter.py`` so the agent can ``python tools/golden_iter.py 8 1e-4 1e-1`` instead of inlining the function — depends on whether the project wants a ``tools/`` directory at all

### DIFF
--- a/program.md
+++ b/program.md
@@ -38,6 +38,25 @@ Each experiment runs on a single GPU. The training script runs for a **fixed tim
 
 **The first run**: Your very first run should always be to establish the baseline, so you will run the training script as is.
 
+**Sweeping a continuous knob (golden-angle sampling)**: When you decide to scan a continuous hyperparameter (LR, weight decay, embedding scale, gate temperature, dropout, …), don't sample uniformly or grid-style — sample from a low-discrepancy golden-ratio sequence on log-space. Plants distribute leaves at the [golden angle 137.5°](https://nph.onlinelibrary.wiley.com/doi/10.1111/nph.16040) because the irrationality of φ is what guarantees no two leaves overlap; the same fact makes the [generalized golden ratio one of the lowest-discrepancy quasi-random sequences](https://extremelearning.com.au/unreasonable-effectiveness-of-quasirandom-sequences/) we know. At the small budgets autoresearch operates on (5–30 candidates per session), it covers the space ~10–30% more efficiently than uniform random and avoids grid search's clustering at endpoints.
+
+```python
+# Drop into a one-shot Python invocation between experiments.
+GOLDEN = 0.61803398875
+def golden_iter(n, lo, hi, seed=0.5):
+    x = seed
+    out = []
+    for _ in range(n):
+        x = (x + GOLDEN) % 1.0
+        out.append(lo * (hi / lo) ** x)        # log-space interpolation
+    return out
+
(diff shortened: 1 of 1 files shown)


svlandeg (commented):

Issue #226, opened 2026-03-12
Principlex


mcclxxxix:
Andrej,

I've been following autoresearch since the March 5 post — 110 changes, 0.862 to 0.858, and the meta-setup iteration where you're optimizing the agent flows more than the repo itself. It's clearly the right paradigm. The ~700 experiment run that found the QKNorm scaler bug confirms that agents can find things humans miss.

What I keep coming back to is the experiment log image you posted. Two columns of DISCARDED entries, a handful of kept improvements, and no way to know why the kept ones worked except that the number went down. The agent found real improvements that transferred from d12 to d24, but nobody can say which of those will transfer to d96, or to a different dataset, or to a different architecture, without re-running the whole search.

I'm proposing a fork called Agent Principlex that adds a visualization-first constraint to the research loop. The core idea comes from PicBreeder research — specifically the discovery that NEAT-evolved CPPNs develop unified factored representations (organized internal circuitry) while SGD-trained networks producing the same outputs develop disorganized patchwork. Same capability, radically different comprehensibility.

Principlex requires the agent to produce visual factored representations as a condition of committing any change. The visualization agenda has four roots:

Engrams — spatial activation maps of what the model has learned to store (adapted from neuroscience / fMRI conventions)
Cuneiform — symbolic decomposition of how features compose during inference (discrete recombinable primitives, like the first writing system)
Directional colors — vector orientation of residual stream representations encoded as hue/saturation/brightness (adapted from diffusion tensor imaging)
Tab2Visual — the principle from diagnostic imaging libraries that every table of numbers must be rendered as an image
The bet: models whose internals show organized factored structure (PicBreeder-style) rather than SGD patchwork will transfer improvements more reliably and be auditable as the codebase scales beyond human review.

Concrete changes to the repo:

principlex.md replaces program.md with a hypothesis-predict-implement-visualize-explain loop
train.py gets lightweight diagnostic hooks (gradient norms, attention entropy, activation stats, residual stream directions) — under 10% overhead
Every committed experiment produces a viz/ directory with four-root visual outputs
Changes that improve val_bpb but produce disorganized visualizations get shelved, not merged
This stays MIT-licensed and credits autoresearch as upstream. The three-file simplicity is preserved. The agent still only touches train.py.

The manifesto and principlex.md are attached. Happy to ship this as a notable fork, open a discussion on the repo, or take whatever form makes sense. The question Principlex asks: can we make comprehension scale alongside capability, before generation 10,205 arrives?

[...]

PR #566, opened 2026-04-30
feat(program): scale-confirm protocol distinguishes transferable wins (related #226)

lonexreb:
## Summary
Adds one paragraph to ``program.md``: every ~20 keeps, run **one** confirming experiment with ``DEPTH`` raised by ~1.5× (e.g. d=8 → d=12), holding everything else fixed. If the improvement persists vs a same-d=12 baseline, the change is *scale-stable* and worth building on. If the improvement vanishes, downgrade the keep to *scale-bound* in ``results.tsv`` and don't use it as a foundation for further keeps.

## Why
``DEPTH = 8`` is one point on the [renormalization-group](https://en.wikipedia.org/wiki/Renormalization_group) flow of this model family. The relevant operators at one scale are different from the relevant operators at another. Empirically, autoresearch's loop has produced wins that don't transfer to larger depths — the [QK-Norm bug discovery at d=8 was a transferable win](https://x.com/karpathy/status/2031135152349524125), but generic LR/scaling/optimizer tweaks frequently aren't. The current loop has no machinery to tell the two apart.

This is the technical core of [#226 (Principlex)](https://github.com/karpathy/autoresearch/issues/226):

> *The agent found real improvements that transferred from d12 to d24, but nobody can say which of those will transfer to d96, or to a different dataset, or to a different architecture, without re-running the whole search.*

The PR doesn't try to predict transferability *in advance* — it just adds a cheap retrospective check (~5% of session budget) so the agent can mark the keeps that survive a single bigger-scale run. Over a session of 100 experiments and ~20 keeps, that's one extra ``d=12`` baseline plus one ``d=12`` confirmation per keep cluster — a few extra runs, not a parallel search.

## Why this is a Feynman move
[Cargo Cult Science](https://calteches.library.caltech.edu/51/2/CargoCult.htm) in one sentence: *surviving the selection criterion isn't the same as being real*. The greedy loop's selection criterion (single-scale ``val_bpb``) doesn't pin down whether a win is local-to-d=8 or genuine. The cure is the one Feynman repeatedly demonstrated: do the experiment that *would* distinguish the two hypotheses, even if it's modestly expensive.

## Composes with everything else
- **PR #560 (noise floor)**: applied to the d=12 baseline so we know whether the d=12 delta is real.
- **PR #555 (Lévy)**: a long-jump experiment that lands a keep is *exactly* the kind of keep that benefits most from scale confirmation.
- **PR #561 (MDL)**: a kept change with negative ``diff_bytes`` (deletion) and positive scale-confirmation is the strongest possible signal — those are the wins worth shouting about.

## Citations
- [Renormalization group (Wikipedia)](https://en.wikipedia.org/wiki/Renormalization_group)
- [Feynman, *Cargo Cult Science*, Caltech 1974](https://calteches.library.caltech.edu/51/2/CargoCult.htm)
- [#226 Principlex (this repo)](https://github.com/karpathy/autoresearch/issues/226)
- ``physics_inspired.md`` §6 in this repo (PR #559)

## Test plan
- [x] Diff is +2 lines, all in one paragraph immediately after the ``Crashes`` paragraph
- [x] Cost-bounded by the protocol itself (~5% of session budget; one d=12 baseline + one confirm per ~20 keeps)
- [x] Cleanly distinguishes scale-stable from scale-bound keeps in ``results.tsv`` so future sessions can preferentially build on the former
- [ ] Optional follow-up: add a ``scale_status`` 6th column to ``results.tsv`` (``stable`` / ``bound`` / ``unverified``) to make the distinction machine-readable


svlandeg (commented):

PR #567, opened 2026-04-30
feat(program): hippocampal-style replay buffer for renewable variation

lonexreb:
## Summary
Adds one paragraph to ``program.md``. Maintain a tiny in-context replay buffer — the last 5 KEEPs and their ``Δval_bpb``. Before each new experiment, with probability ~0.5, pick a buffered KEEP weighted by ``|Δval_bpb|`` and propose a *variant* of it (one knob different, transplanted to a different layer position, scaled to a different model size, combined with another recent KEEP). With probability ~0.5, generate fresh as before.

## Why
Between trials, the hippocampus replays recent trajectories preferentially weighted by reward-prediction error:

> *"Post-learning replay of hippocampal-striatal activity is biased by reward-prediction signals."* — [Ambrose et al., *Nat. Comm.* 2025](https://www.nature.com/articles/s41467-025-65354-2)
>
> *"VTA dopamine signaling drives hippocampal sharp-wave ripples and reverse replay selectively to locations with increased reward."* — [eLife 2024](https://elifesciences.org/articles/99678)

That is — when the brain is between tasks, it doesn't sit idle. It reactivates exactly the recent surprising-win trajectories and re-derives variation from them. The same procedure helps autoresearch: a near-miss KEEP that produced ``0.0007 BPB`` doesn't get forgotten after the next two discards. It gets *repeatedly mined for variants* until it's exhausted.

## What it cures
The current loop's failure mode at scale: the latest result is always the most-salient memory, so a winning idea from 8 experiments ago has no chance of being re-explored unless the agent happens to read back through ``results.tsv`` and *manually* picks it up. Replay makes that happen by default.

## How this composes
- **PR #555 (Lévy)**: replay variants are typically *short* or *medium* jumps from the original KEEP — the heavy-tailed distribution still applies *within* the buffer.
- **PR #558 (IoR)**: even when seeded from the same KEEP, the variant should pick a *different* axis than the last experiment.
- **PR #560 (noise floor)** + **PR #561 (MDL)** + **PR #563 (Metropolis)**: replay only changes *which* experiment to propose; the accept-rule chain is unchanged.

## Cost
Zero compute. The buffer lives in the agent's session context (the agent already reads ``results.tsv`` for context; the replay buffer is just \"the recent KEEP rows\").

## Citations
- [Ambrose et al., post-learning replay biased by reward-prediction signals, *Nat. Comm.* 2025](https://www.nature.com/articles/s41467-025-65354-2)
- [VTA dopamine gates hippocampal replay, *eLife* 2024](https://elifesciences.org/articles/99678)
- [Niv, dopamine RPE hypothesis review, *PNAS* 2009](https://www.pnas.org/doi/10.1073/pnas.1014269108)
- [Posterior vs anterior hippocampus in explore/exploit, *Nat. Comm.* 2020](https://www.nature.com/articles/n41467-020-18864-0)
- ``bioinspired.md`` §4 in this repo (PR #554)

## Test plan
- [x] Diff is +2 lines, all in one paragraph immediately before the ``NEVER STOP`` paragraph
- [x] Zero compute cost; the buffer is the last K rows of ``results.tsv``
- [x] Composes with the existing accept-rule chain without overlap
- [ ] Optional follow-up: track each KEEP's ``predicted_delta`` alongside ``actual_delta`` so the buffer can sample by the *prediction-error* magnitude (the precise dopaminergic gating signal) rather than the absolute Δ. Adds one column to ``results.tsv`` and one line to the analysis loop.


svlandeg (commented):

PR #568, opened 2026-04-30
feat(program): apical-dominance lateral-roots protocol for stall break

lonexreb:
## Summary
Adds one paragraph to ``program.md``. Whenever a KEEP is recorded, the agent also writes a *one-phrase forward-looking note* in the description column (e.g. "high VRAM, may matter at d≥16", "rejected swap, near-miss with ResMLP"). Each note is a **dormant lateral checkpoint** — a fallback hypothesis. **Stall break:** after 8 consecutive DISCARDs, switch to the highest-expected-value-of-information lateral and resume hill-climbing from there.

## Why
Plants concentrate growth at the apical (lead) meristem; [auxin from the lead suppresses lateral buds until the lead is damaged or competition increases](https://en.wikipedia.org/wiki/Apical_dominance), at which point a lateral breaks dormancy and becomes the new lead. The autoresearch analogue: run *one* lead branch plus a portfolio of dormant fallbacks, so a stalled lead doesn't kill the session.

The current loop has no stall-break protocol. Eight discards in a row just become more discards — the agent has nothing in its context to fall back to except "try harder," which empirically just produces more variations on the most-recent failed direction. This PR adds the cheapest possible fallback mechanism: the agent already writes one description per KEEP; this just makes that description forward-looking instead of purely retrospective.

## Cost
Zero — no new TSV column, no new file, no extra runs. Just a richer ``description`` column on KEEP rows. The 8-discard stall threshold is a hyperparameter; if it triggers too rarely, lower it.

## How this composes
- **PR #567 (replay buffer)**: once a lateral is activated, its variants flow naturally into the replay buffer. Replay then mines the new lead the same way it mines any KEEP.
- **PR #560 (noise floor)**: stall detection counts only *real* discards (those above ``2 σ_noise``). Sub-noise wiggles don't count toward the 8-discard threshold.
- **PR #555 (Lévy)**: the act of breaking a lateral is itself a *long jump* in the search graph — discontinuous teleport, not incremental.
- **PR #563 (Metropolis)**: with a non-zero temperature, occasional uphill moves are accepted; this naturally lengthens the time before the stall threshold is hit, and lateral breaks become the *coordinated* escape mechanism rather than the only one.

## Citations
- [Apical dominance (Wikipedia)](https://en.wikipedia.org/wiki/Apical_dominance)
- ``bioinspired.md`` §6 in this repo (PR #554)

## Test plan
- [x] Diff is +2 lines, all in one paragraph immediately after the ``Crashes`` paragraph
- [x] Zero compute cost; only a richer ``description`` column on KEEP rows
- [x] Stall threshold (8 consecutive discards) is a hyperparameter that can be tuned later
- [ ] Optional follow-up: track activated laterals as their own ``status = lateral_break`` rows in ``results.tsv`` so the protocol's effectiveness is auditable


svlandeg (commented):

PR #570, opened 2026-04-30
feat(program): multi-agent section — cuckoo abandonment + octopus loose coupling

lonexreb:
## Summary
Adds a new **Multi-agent setups** section at the bottom of ``program.md``. Two protocols for the ``N > 1`` GPU case:

1. **Cuckoo abandonment**: every K=50 experiments, the agent with the *worst* recent KEEP rate forcibly resets to the *best* agent's current state, then applies a Lévy long-jump. The loop's version of "rotate a researcher off a dead idea." Single-agent analogue is the apical-dominance lateral-roots protocol from PR #568; this is what to do when there's a *peer* to copy from.

2. **Octopus-style loose coupling**: default to *loose* sync. Each agent runs locally with strong defaults and only synchronizes via the shared ``results.tsv`` + global pheromone table once per ~20 experiments. Tight blocking sync is the single largest reason multi-agent setups underperform ``N`` solo agents — every sync is ``N-1`` idle GPUs.

## Why
The repo already supports the multi-agent shape: ``program.md`` mentions branches like ``autoresearch/<tag>-gpu0`` and ``autoresearch/<tag>-gpu1``. But the current loop has no coordination protocol — multi-agent runs are just N copies of the single-agent loop, with no information flow between them. The two failure modes:

- **No information flow**: the agents end up duplicating exploration. A KEEP one agent finds at experiment 30 isn't visible to peers running on the same idea, so they re-discover it 60 experiments later.
- **Tight synchronization**: any naive fix (broadcast every result to every agent) introduces blocking syncs and idle GPUs.

The cuckoo + octopus combination is the cheapest setup that gets information flow without the blocking-sync tax: agents run independent, only sync at K=20-50 cadence, and the worst agent gets a forced reseed periodically.

## Why "single-agent equivalent already exists"
PR #568 (apical-dominance lateral-roots) gives the single-agent stall break: when the lead branch stalls, switch to a saved lateral. Cuckoo is the multi-agent generalization — instead of saving laterals on the same branch, the *peers* are the laterals, and "stall" is now relative ("worst peer" rather than "absolute discard streak"). The two protocols are the same idea at different cardinalities.

## Citations
- [Cuckoo Search (Wikipedia)](https://en.wikipedia.org/wiki/Cuckoo_search)
- [Cuckoo Search applied to NoC mapping with Lévy flight, IEEE 2021](https://ieeexplore.ieee.org/document/9570345/)
- ``bioinspired.md`` §7 and §9 in this repo (PR #554)

## Independence and composition
- Pure prompt addition. No code, no infra changes to the existing single-agent ``LOOP FOREVER`` block above it.
- The single-agent loop (with all the patterns from PRs #555/#558/#560/#561/#562/#563/#564/#565/#566/#567/#568/#569) is unchanged. This section is *additive*: read it only when ``N > 1``.

## Test plan
- [x] Diff is +8 lines, all in one new section at the bottom of the file
- [x] No change to the single-agent loop semantics
- [x] Coordination protocols specified at K=50 / K=20 cadences (tunable hyperparameters)
- [ ] Optional follow-up: a tiny ``tools/swarm_coord.py`` script that watches all ``autoresearch/<tag>-gpuN`` branches and issues the cuckoo reset; only worth writing if/when someone runs the multi-agent shape in practice.


svlandeg (commented):

PR #571, opened 2026-04-30
feat(train): log loss_std_last_100 as honest-loss stability signal

lonexreb:
## Summary
Adds one extra line to the final summary printed by ``train.py``:

```
loss_std_last_100:0.0624
```

It's the population stdev of ``train_loss`` over the last 100 training steps, captured via a 100-element ``collections.deque``. No GPU work, no impact on the metric.

## Why
Two runs at the same ``val_bpb`` but very different loss volatility are not equivalent. The current loop has no signal that distinguishes them. ``val_bpb`` is the *energy*; ``loss_std_last_100`` is one of the *entropy*-like degrees of freedom that change the free energy without changing the energy (cf. ``physics_inspired.md`` §7, PR #559).

Empirically: a change that drops ``val_bpb`` but doubles ``loss_std_last_100`` is trading robustness for fit. It is likely to underperform when scaled to larger ``DEPTH`` (cf. the renormalization-group scale-confirm protocol in PR #566) or to a different dataset.

The existing fast-fail check (``math.isnan(loss) or loss > 100``, line 577) only catches *catastrophic* divergence. ``loss_std_last_100`` catches the in-between case: training that didn't crash but ran with unstable dynamics throughout.

## Verified locally
Synthetic stable vs volatile runs at *identical* final loss:

```
stable:   loss_std_last_100 = 0.0624
volatile: loss_std_last_100 = 0.1306    (2.1x)
```

Same final loss; the stability signal cleanly separates them.

## Implementation
Minimal diff. Three additions in ``train.py``:

1. ``import statistics`` and ``from collections import deque`` (both stdlib, no new deps).
2. ``recent_losses = deque(maxlen=100)`` initialized before the training loop.
3. ``recent_losses.append(train_loss_f)`` immediately after the existing ``train_loss_f = train_loss.item()``.
4. ``print(f"loss_std_last_100:{loss_std_last_100:.4f}")`` appended to the existing final-summary block.

No change to the training loop's hot path beyond a single ``deque.append`` per step (O(1)). No CUDA work, no extra sync.

## How it composes
- **PR #560 (noise floor)** + **PR #561 (MDL)** + **PR #563 (Metropolis)** filter the *accept rule*. This PR adds an *honest-loss consistency check* the agent can read after a KEEP — orthogonal to whether the change was accepted.
- **PR #566 (RG scale-confirm)**: a KEEP with high ``loss_std_last_100`` is exactly the case that benefits most from scale confirmation before being built upon.

## Citations
- ``physics_inspired.md`` §7 (PR #559) in this repo

## Test plan
- [x] Synthetic stable vs volatile runs at identical final loss → 2.1× separation in ``loss_std_last_100``
- [x] Imports are stdlib only (``statistics``, ``collections.deque``)
- [x] No GPU work, no extra ``torch.cuda.synchronize`` calls
- [x] Diff is +10 lines total: 2 imports, 1 ring buffer init, 1 append, 1 final-summary print, plus the explanatory comment
- [ ] If merged: existing ``results.tsv`` keeps its 5-column shape unchanged. Optional follow-up: add ``loss_std`` as a 6th column for cross-experiment auditability of the stability signal.

### DIFF
--- a/train.py
+++ b/train.py
@@ -10,7 +10,9 @@ os.environ["HF_HUB_DISABLE_PROGRESS_BARS"] = "1"
 
 import gc
 import math
+import statistics
 import time
+from collections import deque
 from dataclasses import dataclass, asdict
 
 import torch
@@ -539,6 +541,11 @@ t_start_training = time.time()
 smooth_train_loss = 0
 total_training_time = 0
 step = 0
+# Honest-loss signal: ring buffer of recent train_loss values; reported in the
+# final summary as `loss_std_last_100`. Two runs at the same val_bpb but very
+# different loss-volatility are not equivalent — high volatility usually
+# means the win is fragile and won't transfer (cf. physics_inspired.md §7).
+recent_losses = deque(maxlen=100)
 
 while True:
     torch.cuda.synchronize()
@@ -565,6 +572,7 @@ while True:
     model.zero_grad(set_to_none=True)
 
     train_loss_f = train_loss.item()
+    recent_losses.append(train_loss_f)
 
     # Fast fail: abort if loss is exploding or NaN
     if math.isnan(train_loss_f) or train_loss_f > 100:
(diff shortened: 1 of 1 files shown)


svlandeg (commented):

PR #587, opened 2026-05-11
Add EMI manuscript DOCX deliverables.

MehmetDEU:
Package binary manuscript drafts separately so document-format checks and distribution can be reviewed independently from source markdown edits.


svlandeg (commented):
I assume that this PR and your other ones were meant to be PR'd against your fork instead of upstream `master`?

PR #588, opened 2026-05-11
Bump PyTorch to 2.11.0 (fixes silent expandable_segments bug in 2.9.1)

srosro:
## Summary

**Reopening with a different fix.** The original commit on this branch renamed `PYTORCH_ALLOC_CONF` → `PYTORCH_CUDA_ALLOC_CONF` as a "typo fix" — @svlandeg's [comment on #510](https://github.com/karpathy/autoresearch/pull/510) correctly flagged that as wrong: `PYTORCH_ALLOC_CONF` is the new unified env var name per PyTorch's deprecation direction, and `train.py:8` is forward-looking and correct.

The real bug is in PyTorch 2.9.1: it silently ignores `expandable_segments:True` set via the new `PYTORCH_ALLOC_CONF` for CUDA, only honoring the deprecated alias. Verified fixed in PyTorch 2.11.0. This PR bumps the pin:

```diff
- "torch==2.9.1",
+ "torch==2.11.0",
```

(plus regenerated `uv.lock`).

No `train.py` change needed — the upstream env-var line was always correct.

## Measurements on RTX 3090 (24 GB), `DEVICE_BATCH_SIZE=64`, upstream `train.py` as-is

| PyTorch | `expandable_segments` actually enabled? | Outcome |
|---|---|---|
| 2.9.1 | ❌ False (silent, despite the env var being set) | OOM at batch=64 (1.91 GiB stuck reserved-unallocated) |
| **2.11.0** | ✅ True | Fits at peak 22.8 GiB, `val_bpb=1.2530` |

## Relationship to other open PRs

- **#504** (@ademeure) bumps torch to 2.11 as part of a larger Blackwell + FlexAttention + SDPA + MFU change (+290/-205 across 3 files). If you want the comprehensive fix, that's the more capable PR — my data above also supports landing it.
- **#510** (@mendentm) and **#546** (@lonexreb) rename the env var to the deprecated alias as a workaround. Both work on 2.9.1 but trigger PyTorch's "deprecated, use PYTORCH_ALLOC_CONF instead" warning every run, and are unnecessary once you're on 2.11+.

This PR is the minimal one-line scope: just the version bump, no other changes. Useful if you want to unblock the batch=64 OOM specifically without absorbing the FlexAttention/SDPA delta.

## Note re: the commit history on this branch

Two commits — the original wrong-framed rename + a second commit that reverts it and bumps `pyproject.toml`/`uv.lock`. Net diff against `master` is just the version bump. Happy to squash or rebase.

## Test plan

- [x] Verified on RTX 3090, 24 GB VRAM, Linux + driver 580 / CUDA 13
- [x] Before bump: 2.9.1, upstream code, `uv run train.py` at batch=64 → `torch.OutOfMemoryError` with 1.91 GiB reserved-unallocated
- [x] After bump: 2.11.0, upstream code, `uv run train.py` at batch=64 → completes the 5-min budget, `val_bpb=1.2530`, peak 22.8 GiB
- [x] PyTorch 2.11 fix verified directly via `torch.cuda.memory._snapshot()['allocator_settings']['expandable_segments']` returning `True` with the new env var name

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ dependencies = [
     "requests>=2.32.0",
     "rustbpe>=0.1.0",
     "tiktoken>=0.11.0",
-    "torch==2.9.1",
+    "torch==2.11.0",
 ]
 
 [tool.uv.sources]
--- a/train.py
+++ b/train.py
@@ -5,7 +5,7 @@ Usage: uv run train.py
 """
 
 import os
-os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
+os.environ["PYTORCH_ALLOC_CONF"] = "expandable_segments:True"
 os.environ["HF_HUB_DISABLE_PROGRESS_BARS"] = "1"
 
 import gc
--- a/train.py
+++ b/train.py
@@ -5,7 +5,7 @@ Usage: uv run train.py
 """
 
 import os
-os.environ["PYTORCH_ALLOC_CONF"] = "expandable_segments:True"
+os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
 os.environ["HF_HUB_DISABLE_PROGRESS_BARS"] = "1"
 
 import gc
--- a/uv.lock
+++ b/uv.lock
@@ -69,7 +69,7 @@ requires-dist = [
     { name = "requests", specifier = ">=2.32.0" },
     { name = "rustbpe", specifier = ">=0.1.0" },
     { name = "tiktoken", specifier = ">=0.11.0" },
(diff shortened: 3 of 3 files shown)


svlandeg (approved):
Thanks for the digging @srosro! 

I suggest we squash-merge this PR if Andrej is happy upgrading PyTorch to 2.11.0.

svlandeg:
> The real bug is in PyTorch 2.9.1: it silently ignores expandable_segments:True set via the new PYTORCH_ALLOC_CONF for CUDA, only honoring the deprecated alias.

~~Is there a bug report or PR you can link to that actually confirms this was a bug in PyTorch 2.9.1? I can't find it.~~

I see you mentioned this one on another PR: https://github.com/pytorch/pytorch/issues/183387, but that's yours, I would expect there is some other trace of this bug to be found in the PyTorch repo 🤔

PR #614, opened 2026-06-21
fix(train): reduce rotary embedding sequence length to sequence_len

### LINKED ISSUE #6: Rotary embeddings pre-allocate 10x more memory than needed
## Problem

In \GPT.__init__()\, rotary embeddings are pre-computed for 10x the actual sequence length:

\\\python
self.rotary_seq_len = config.sequence_len * 10  # 20480 positions
\\\

Since \MAX_SEQ_LEN = 2048\ and the training/eval code never exceeds this length, the 10x factor wastes GPU memory on unused cos/sin buffers (10x larger than necessary in bfloat16).

## Proposed Fix

Set \self.rotary_seq_len = config.sequence_len\. One-line change that reduces the rotary buffer memory by 90% with zero impact on training or evaluation (both are bounded by \MAX_SEQ_LEN\).

I'll submit a PR for this.

aniruddhaadak80:
Resolves #6. Pre-allocating 10x more sequence length memory than needed for rotary embeddings in GPT.__init__() wastes GPU memory. This change sets self.rotary_seq_len to config.sequence_len, reducing rotary buffer memory by 90% with zero impact on training/evaluation (which are bounded by MAX_SEQ_LEN).

### DIFF
--- a/train.py
+++ b/train.py
@@ -141,7 +141,7 @@ class GPT(nn.Module):
             for i in range(config.n_layer) if has_ve(i, config.n_layer)
         })
         # Rotary embeddings
-        self.rotary_seq_len = config.sequence_len * 10
+        self.rotary_seq_len = config.sequence_len
         cos, sin = self._precompute_rotary_embeddings(self.rotary_seq_len, head_dim)
         self.register_buffer("cos", cos, persistent=False)
         self.register_buffer("sin", sin, persistent=False)


svlandeg (commented):
Closing, duplicate of https://github.com/karpathy/autoresearch/pull/78 (and maybe others), which could have been easily spotted as they are linked from https://github.com/karpathy/autoresearch/issues/6.

PR #615, opened 2026-06-21
fix(train): dynamically detect GPU capability and look up peak FLOPS for MFU calculation

### LINKED ISSUE #5: MFU calculation is hardcoded to H100 peak FLOPS

aniruddhaadak80:
Resolves #5. The Model FLOPS Utilization (MFU) calculation previously assumed hardcoded H100 peak BF16 Tensor FLOPS (989.5 TFLOPS). This change detects the active GPU device name using torch.cuda.get_device_name(0), and sets the theoretical peak FLOPS using a lookup table for common GPUs (H100, A100, L40S, L4, RTX 4090, RTX 3090, A10G, etc.), falling back to the H100 peak value for unknown devices. This ensures MFU is accurate and comparable across different hardware environments.

### DIFF
--- a/train.py
+++ b/train.py
@@ -460,7 +460,42 @@ torch.cuda.manual_seed(42)
 torch.set_float32_matmul_precision("high")
 device = torch.device("cuda")
 autocast_ctx = torch.amp.autocast(device_type="cuda", dtype=torch.bfloat16)
-H100_BF16_PEAK_FLOPS = 989.5e12
+def get_device_peak_flops(device_name: str) -> float:
+    name = device_name.lower()
+    if "h100" in name:
+        return 989.5e12
+    elif "a100" in name:
+        return 312.0e12
+    elif "4090" in name:
+        return 165.0e12
+    elif "3090" in name:
+        return 142.0e12
+    elif "a6000" in name:
+        return 154.0e12
+    elif "l40s" in name:
+        return 366.0e12
+    elif "l40" in name:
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
`nanochat` has this spelled out in more detail, cf https://github.com/karpathy/nanochat/blob/dc54a1a3077cab11d68fac4c5d1cd5c51f5d8c7a/nanochat/common.py#L227, but I think the idea for `autoresearch` was to provide a more minimal training loop to allow for quick experiments. Users can fork and adjust this as they please.

PR #619, opened 2026-06-21
feat(novelty): add experiment novelty score checker to prevent redundant runs

### LINKED ISSUE #613: Feature: Experiment Novelty Score to Reduce Repeated Search Patterns
### Is your feature request related to a problem? Please describe.

As the number of experiments grows, the agent may revisit similar regions of the search space despite maintaining a history of previous runs.

While repeated validation is sometimes useful, it can also reduce exploration efficiency by spending compute on ideas that are only minor variations of previously rejected experiments.

### Describe the solution you'd like

Introduce an optional experiment novelty score.

Before launching a new experiment, compare the proposed modification against recent accepted and rejected experiments.

The score could consider:

* Hyperparameter similarity
* Architectural similarity
* Optimization strategy similarity

Low-novelty experiments could be flagged in the experiment log, allowing users to better understand whether the agent is genuinely exploring new directions or repeatedly searching nearby solutions.

### Describe alternatives you've considered

Users can manually inspect experiment histories, but this becomes increasingly difficult as runs scale to dozens or hundreds of experiments.

### Additional context

This feature would not change the keep-or-revert workflow. Instead, it would provide additional visibility into search diversity and exploration behavior during long autoresearch runs.

aniruddhaadak80:
Resolves #613. During long autonomous search runs, the AI agent can repeatedly search the same regions of the hyperparameter or architectural search space. This PR introduces a standalone novelty.py utility. Before launching a new experiment, the agent can call novelty.py with the proposed experiment description. The utility parses results.tsv and computes a similarity score against previous runs using Jaccard word-level overlap and character-level sequence matching. It flags repeated patterns or low-novelty proposals, improving exploration efficiency without changing the core keep-or-revert workflow.

### DIFF
--- a/novelty.py
+++ b/novelty.py
@@ -1,0 +1,150 @@
+#!/usr/bin/env python3
+"""
+Experiment Novelty Score Checker for autoresearch.
+Compares a proposed experiment description against previous runs in results.tsv
+to prevent the agent from repeating the same hyperparameter/architectural changes.
+
+Usage:
+    python novelty.py "description of the proposed experiment"
+"""
+
+import sys
+import os
+import difflib
+
+def get_past_experiments(tsv_path="results.tsv"):
+    """Reads results.tsv and returns a list of past experiment dicts."""
+    if not os.path.exists(tsv_path):
+        return []
+    
+    experiments = []
+    try:
+        with open(tsv_path, "r", encoding="utf-8") as f:
+            lines = f.readlines()
+            if len(lines) <= 1:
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
I personally think that an addition like this one is something all users can experiment with on their own fork, and doesn't necessarily have to be PR'd against upstream `master`. The idea of the `autoresearch` repo is for you to fork it and play around with it to work out what works best in your setting :-)

PR #620, opened 2026-06-22
[codex] fix: count tokenizer bytes from raw ranks

### LINKED ISSUE #384: Bug: BPB metric inflated by UTF-8 replacement characters in token byte count

chaochaoweb3:
## Summary
- build BPB token byte lengths from raw BPE mergeable ranks instead of decode/re-encode text
- keep special tokens at byte length 0 so they stay excluded from BPB byte totals
- add a small helper for the byte-length table used during tokenizer preparation

Fixes #384.

## Validation
- `python3 -m py_compile prepare.py`
- AST-extracted `build_token_byte_lengths` self-check with invalid UTF-8 raw byte token and special-token gap
- `git diff --check`

Full tokenizer training was not run; this change only affects the generated token byte lookup logic.


svlandeg (commented):
There's already several PRs open for that particular issue - closing as duplicate. Please feel free to review the existing PRs instead! 🙏

